PR #11, opened 2016-07-21
Fix finding close unused pokestops

jabbink:


jabbink:
@vmarchaud this possibly fixed your problem

PR #15, opened 2016-07-21
Simplified

SwipeX:
simplified a few expressions to val from var.
inserted a property expression in place of a getter.
introduced a lambda.


jabbink:
Thanks, merged

Issue #13, opened 2016-07-21
NoSuchElementException


zeladada:
`Exception in thread "Thread-34" java.util.NoSuchElementException: List is empty.
	at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:180)
	at ink.abb.pogo.scraper.MainKt.processMapObjects(Main.kt:181)
	at ink.abb.pogo.scraper.MainKt$main$$inlined$fixedRateTimer$lambda$lambda$1.invoke(Main.kt:126)
	at ink.abb.pogo.scraper.MainKt$main$$inlined$fixedRateTimer$lambda$lambda$1.invoke(Main.kt)
	at kotlin.concurrent.ThreadsKt$thread$thread$1.run(Thread.kt:18)`
The script works without errors for about 30 seconds but afterwards this exception is printed every time processMapObjects is called when no catchable pokemon are near. The program continues to run without problems despite the exception being thrown.

PR #26, opened 2016-07-21
Task framework added

SwipeX:
Each larger block of code can be simplified into a Task. Using the context object as a simpleton to share information.


SwipeX:
previously the call was being made twice between size and first.  If there was only one pokemon and it was removed between these calls, it would cause the error as noted in #13

PR #36, opened 2016-07-21
Added option to change number of useless items and show Rewards Information

CosmoRing:
Now you can use the config properties to change the number of items the bot should keep before starting to drop the excess. Possible to not drop an item by using -1


falko100:
+1 :+1:

shacoune:
OMG dude can't believe ! Rockstar ! (Y)

Crackwizard:
U are the best!

src/main/kotlin/ink/abb/pogo/scraper/Main.kt
```diff
@@ -118,6 +118,23 @@ fun main(args: Array<String>) {
         println("No item drop policy specified, defaulting to $shouldDropItems")
     }
 
+    if(shouldDropItems){
+        HashMap(uselessItems).forEach {
+            val itemName = it.key.name.toLowerCase()
+
+            try {
+                val newValue = properties.getProperty(itemName).toInt()
```
sirsean: What's the behavior of this if the property is not specified in the config file?

Will it give an NPE, or default to 0 (which would fail the `newValue < -1` check below ... and drop all your items of the omitted type).

I think we ought to try to make sure an unchanged config file (ie, omitting all these values) has the behavior of using the default values in the code.

CosmoRing: It will use the defaults already in uselessItems.

The val uselessItems is an mutableMapOf so you can update the map. The map is in 332 and already has the default values. 

If some properties aren't in the config file or aren't valid ( greater then -1 ) it will inform about the default used.

CosmoRing: It will use the defaults already in uselessItems.

The val uselessItems is an mutableMapOf so you can update the map. The map is in 332 and already has the default values.

If some properties aren't in the config file or aren't valid it will inform about the default used.

sirsean:
This LGTM once those slight typing errors are fixed. Should merge it quick!

sirsean:
👍 :shipit:

CosmoRing:
@jabbink Should i fix this or there isn't a need for this in the bot?

pr0ves:
I think this is a nice feature, to add it you'd just need to adapt it to the new structure

jabbink:
Love the addition and I'd love it if you rebased this (probably less work take the diff of this and `master` and manually copy the changes in and then submit it as one squashed commit).

Check the `Settings.kt` class, it contains a way to get settings with a default value converted to some type and warns users with old configs to add these values

CosmoRing:
Ok, i can do it later but can only start doing it +-4h from now. So if someone is in a rush you can do it

jabbink:
Not in a rush. Will release the compiled 0.3.0 then, can always add this

CosmoRing:
Ok i think it's ready. Can someone take a look?

jabbink:
As reference: closes #85

src/main/kotlin/ink/abb/pogo/scraper/tasks/WalkToUnusedPokestop.kt
```diff
@@ -33,6 +33,8 @@ class WalkToUnusedPokestop(val sortedPokestops: List<Pokestop>) : Task {
         }
 
         if (nearestUnused.size > 0) {
+            if (settings.shouldDisplayWalkingToNearestUnused)
+                Log.normal("Walking to pokestop \"${nearestUnused.first().details.name}\"")
```
jabbink: FYI: Adds one HTTPS request per log

CosmoRing: Ok, didn't know that. Thanks for the info

jabbink:
Merge incoming, no need to resolve conflicts

PR #38, opened 2016-07-21
Split settings and bot code into their own classes.

apottere:
I took the liberty of separating the 3 components (initialization, settings, and botting) up so the code is easier to understand.

Functional changes included in this PR:
* `base64_password` setting is used if `password` is missing, allowing your password to be base64 encoded.  This doesn't mean it's secure, but at least it's not in plaintext so it can't be read over your shoulder.
* ~~Added pokemon CP transfer threshold as a setting.~~


falko100:
:-1:  I don't want the pass encoded. Maybe give the option for encoded password ?

apottere:
Yeah, I can do that.  Working through the merge conflicts right now.

apottere:
Ok, I added a `base64_password` setting that is checked if `password` is empty.

jabbink:
Thanks, looks good. Now I'll have to see how to efficiently merge this with the current `master` and #39 :P

jabbink:
Should be merged in https://github.com/jabbink/PokemonGoBot/tree/pr/38-39

Can't test as the servers are offline

apottere:
Thanks!  I can help merge stuff in too, I realize it probably caused a lot of conflicts.

PR #47, opened 2016-07-22
Walk to Wild Pokemon and new JAVA API PTC class

davidmann4:
* bot walks to wild pokemons
* classname change from PTCLogin to PtcLogin
* bot does not walk away until he catched all bots on location

edit: ups commited my values for item drops - well will be changed with configurable drops anyways

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Main.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Main.kt
@@ -23,7 +23,7 @@ import com.pokegoapi.api.map.fort.Pokestop
 import com.pokegoapi.api.player.PlayerProfile
 import com.pokegoapi.api.pokemon.Pokemon
 import com.pokegoapi.auth.GoogleLogin
-import com.pokegoapi.auth.PTCLogin
+import com.pokegoapi.auth.PtcLogin
 import okhttp3.OkHttpClient
 import ink.abb.pogo.scraper.Context
 import ink.abb.pogo.scraper.tasks.Release
@@ -123,7 +123,7 @@ fun main(args: Array<String>) {
     if (username.contains('@')) {
         auth = GoogleLogin(http).login(username, properties.getProperty("password"))
     } else {
-        auth = PTCLogin(http).login(username, properties.getProperty("password"))
+        auth = PtcLogin(http).login(username, properties.getProperty("password"))
     }
     println("Logged in as ${properties.getProperty("username")}")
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/ink/abb/pogo/scraper/Main.kt
```diff
@@ -313,16 +332,16 @@ fun processMapObjects(api: PokemonGo, pokestops: MutableCollection<Pokestop>) {
 }
 
 val uselessItems = mapOf(
-        Pair(ItemId.ITEM_REVIVE, 8),
+        Pair(ItemId.ITEM_REVIVE, 30),
         Pair(ItemId.ITEM_MAX_REVIVE, 8),
         Pair(ItemId.ITEM_POTION, 5),
-        Pair(ItemId.ITEM_SUPER_POTION, 5),
-        Pair(ItemId.ITEM_HYPER_POTION, 5),
+        Pair(ItemId.ITEM_SUPER_POTION, 50),
```
davidmann4: ups :)

Crackwizard:
nice

jabbink:
Uses only WildPokemon, also you shouldn't walk towards the WildPokemon (they are catchable already).

davidmann4:
Hmm as far as I understood the names:

wildPokemons == pokemons on the map visible in client
catchablePokemon == in range to engage

so my script walks to wild pokemons to make them catchable

davidmann4:
also - quick question: I did not found any loop for throwing pokeballs. Sometimes pokemons escape the first throw. Would this result in a failed attempt atm?

jabbink:
WildPokemon are engageable. Also I slightly changed the code and will merge it in a second.

davidmann4:
cool - thx. can you explain the difference between:

NearbyPokemon
WildPokemon
catchablePokemon 

also the return statement after a sucessfull catch is important IMO - pokestops are nice but catching pokemons should be number one prio.

PS: always nice to see projects which accept/interact with PR :+1:

jabbink:
Pokemon were always number 1 prio in this bot.

NearbyPokemon -> 3 footsteps in the app, need to walk to catch; client does not know exact coordinates

WildPokemon -> not directly beneath you, but server allows capturing it and the client determines itself when to show these to you (not all are shown at all times); 1-2 footsteps

CatchablePokemon -> you're basically standing on them and the official client shows them, like WildPokemon directly catchable.

davidmann4:
so why we do use CatchablePokemon in the loop? sounds like

CatchablePokemon.count() < WildPokemon.count()

jabbink:
Uhm, can't seem to find which line you mean.

davidmann4:
https://github.com/jabbink/PokemonGoBot/blob/master/src/main/kotlin/ink/abb/pogo/scraper/Main.kt#L213

jabbink:
That line returns `WildPokemon` and `MapPokemon`. Neither require walking.

Also a huge refactoring is incoming which should make the code a lot more readable.

PR #65, opened 2016-07-22
Fixed wrong level progress percentage

### LINKED ISSUE #22: % of XP to level up broken
from 90-ish % to 40-ish %.

I think the error is there [Main.kt#L.295](https://github.com/jabbink/PokemonGoBot/blob/master/src/main/kotlin/ink/abb/pogo/scraper/Main.kt#L295).

Will do a PR if I have time later ;)

benestr:
Using the required xp list rom the decoded protobuf file(https://www.reddit.com/r/pokemongodev/comments/4t59t1/decoded_game_master_protobuf_file_v01_all_pokemon/d5f2rk8)

Fixes #22

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/UpdateProfile.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/UpdateProfile.kt
@@ -15,11 +15,16 @@ import ink.abb.pogo.scraper.Task
 import java.text.DecimalFormat
 
 class UpdateProfile : Task {
+
+    val requiredXp = arrayOf(0, 1000, 3000, 6000, 10000, 15000, 21000, 28000, 36000, 45000, 55000, 65000, 75000,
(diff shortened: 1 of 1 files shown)


jabbink:
What do the numbers in stats.nextLevelXp represent anyways then?

Because if those numbers somehow contain the required information as well, I'd rather use this dynamic provided info instead of some static list of numbers that might change in the future and is extracted from the apk.

benestr:
prevLevelXp is the xp required for the level before the current one and nextLevelXp is xp required for the next level up. 
So if I'm level 13 prevLevelUp is the xp required to reach level 12(65000) and nextLevelUp is the xp required for level 14(85000). No idea why they would do it that way...
So as far as I understand there is no way to get the xp that was needed for the current level (75000) without a list, it's not guaranteed to be in the middle. Considering the info is in the apk the app probably uses the list, too.

I doubt they would change the xp requirements anyway. Players would suddenly jump levels, not to mention the level up rewards.

apottere:
+1 from me - if/when the data is updated, we can always change it back.  If the levels are included in the APK, it's almost guaranteed that they aren't changing.

PR #69, opened 2016-07-22
Add color in console

dommerq:
Add some pretty colors so it's easier to read when you catch some pokemon (green) or when something when wrong (red)

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Log.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Log.kt
@@ -1,3 +1,6 @@
+import java.text.SimpleDateFormat
+import java.util.*
+
 class Log {
 
     companion object {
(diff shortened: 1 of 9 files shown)


alexvndre:
👍

apottere:
Could we use something like this instead of rolling our own log? https://github.com/Tuxdude/logback-colorizer/blob/master/README.md

dommerq:
Looks overkill to me, I think there's too much config on this ¯\_(ツ)_/¯

RaeneDev:
gotta make things perty. 
If nothing has been added/researched i can look into something for this in a couple hours. I was possibly thinking of making a GUI to add a launch button, change config settings, and an option to hide output such as item drops but I mostly do web languages so I'm not sure how well or even if I could do this.

jabbink:
Too be honest, using slf4j with logback sounds better imo, but I have to agree that this is an overkill for to add now and in case we ever want to switch, replacing this with slf4j should still be easy to do.

jabbink:
Also my console apparently doesn't get resets.

image

dommerq:
wow that's weird. Reset works because you get the normal grey color after but I don't know what those characters are :/ 
Will look and update this PR.

jabbink:
While you're at it, could you add logging of the time?

jabbink:
Reset works fine in the IntelliJ console by the way.

Probably some bug with pantheon-terminal + zsh.

dommerq:
Can't test pantheon-terminal, I'm on OSX. (using zsh on OSX, it works :/ )

I'm gonna add the time yep

apottere:
@RaeneDev I'm looking into adding a web frontend too.

jabbink:
Too late, but I notice the namespace of the logging thing is `Log`

Will change that tomorrow.

Issue #45, opened 2016-07-22
Recommendation: Compute IV and Auto Transfer <%


xenxes:
Use formula via https://docs.google.com/spreadsheets/d/1wbtIc33K45iU1ScUnkB0PlslJ-eLaJlSZY47sPME2Uk/htmlview?usp=sharing&sle=true to calculate IV based on CP, HP, dust price of captured pokemon and auto-transfer those below a set threshold.

pr0ves:
I did something similar in #39 but I just used the CP. 
The dust price is the dust needed to level the pokemon up? 
Anyway all the characteristics of a pokemon that we can use are defined here https://github.com/AeonLucid/POGOProtos/blob/master/src/POGOProtos/Data/PokemonData.proto where stamina and stamina_max should be the HP, but i don't see the dust price.

jabbink:
What does IV stand for?

Also stuff like the Dust price is available in the GAME_MASTER protobuf, which will be incorporated in the Java API soon'ish, so it would be futile to store them in this repo as well then.

mroge009:
I added IV calculation with options in the properties to set minIV.

config.properties
```
#Minimum IV to keep a pokemon
transfer_IV_threshold=90
```

Bot.kt
```
api.pokebank.pokemons.map {
            var IV = ((it.individualStamina + it.individualAttack + it.individualDefense) * 100) / 45
            "Have ${it.pokemonId.name} (${it.nickname}) with ${it.cp} CP and IV ${IV}"
        }.forEach { println(it) }
```

Settings.kt
`val transferIVThreshold = getPropertyIfSet("Minimum IV to keep a pokemon", "transfer_IV_threshold", 80, String::toInt)`

Releasepokemon.kt
```
val minIV = settings.transferIVThreshold

        groupedPokemon.forEach {
            val sorted = it.value.sortedByDescending { it.cp }
            for ((index, pokemon) in sorted.withIndex()) {
                var IV = ((pokemon.individualStamina + pokemon.individualAttack + pokemon.individualDefense) * 100) / 45
                // never transfer highest rated Pokemon
                // never transfer > maxCP, unless set in obligatoryTransfer
                // stop releasing when pokemon is set in ignoredPokemon
                when (pokemon.pokemonId.name) {
                    "RATTATA" -> if (index > pokemon.candy / 25) {
[...]

PR #77, opened 2016-07-22
Changed CP threshold with IV percentage threshold

pr0ves:
Did as asked in #45

### DIFF
--- a/config.properties.template
+++ b/config.properties.template
@@ -14,8 +14,9 @@ drop_items=true
 speed=2.778
 # should the bot auto transfer duplicate pokemon
 autotransfer=false
-# Minimum CP to keep a pokemon
-transfer_cp_threshold=400
+# Minimum IV percentage to keep a pokemon
+# between 0 and 100, suggested 80
+transfer_iv_threshold=80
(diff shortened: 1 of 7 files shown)


jabbink:
:+1:

can you make it compatible with #69 (colors) though?

pr0ves:
As soon as i get home.

vohphuong:
I got when building

Cannot access 'username': it is 'private' in 'PlayerProfile'

and many lines

pr0ves:
@pvase666 I'm fixing it right now

pr0ves:
@pvase666 it's strange that you have that problem because username is accessed through the getter.
Also i think you should post this as issue not here, create an issue and post all of the error not just one line.

pr0ves:
You can merge now.

pr0ves:
Added only the count of pokemon, can't get the number of items because there's no function in Bag class to return the item list

pr0ves:
There's a problem with the stardust, it doesn't get updated after the bot catches a pokemon.
It's probably set the first time and never updated. Is this a pokegoapi problem or ours?

PR #115, opened 2016-07-23
Softban detection

meMo-Minsk:
When bot stopped at Pokestop and find 0 XP and 0 items then you are probably soft banned.
Trying to wait for some minutes...


jabbink:
See, https://github.com/jabbink/PokemonGoBot/pull/124/commits/f75022cc3d4965ebe978973d178d814006d930fc#diff-5c5bc47946b29666c9e46807b63efff1R67

doesn't work

PR #117, opened 2016-07-23
Added option to set prefered pokemon when having multiple that can be catched

shilch:
When having multiple pokemon that can be catched, the bot will prefer these in the list.
Also updated authors.md


jabbink:
I don't think this is really that useful, it doesn't often happen that there are multiple pokemon on one spot, and even then it will quickly catch all of them.

shilch:
Ok did not know that.

PR #123, opened 2016-07-23
hatch eggs

jabbink:


KyleBoyer:
Will this put the highest KM egg in the disposable incubator and the lowest in the reusable? This is how you'd get the most usage out of the incubators.

jabbink:
Nope, while coding this I thought I could be lazy as you always have the infinite incubator and the bot walks so much that on all my bot accounts I didn't have any eggs at all after 3-4 hours. But this can be fixed, yes.

PR #134, opened 2016-07-23
Only transfer pokemon when both IV and CP are lower than treshold

### LINKED ISSUE #126: Behaviour of ReleasePokemon
Right now the bot transfers a pokemon if it has iv < maxIV OR if cp<maxCP.
Shouldn't it be  iv < maxIV AND if cp<maxCP?

e-fisher:
Only transfer pokemon when both IV and CP are lower than threshold, resolves #126


pr0ves:
Done in #135

jabbink:
Also this didn't allow setting `-1` as value to not care about either CP or IV.

PR #147, opened 2016-07-23
Moved the shouldAutoTransfer check before starting the thread

pr0ves:
As title.


pr0ves:
Added feature to save the token directly to config.properties. The problem is that it shuffles the orders of the properties when writing. A solution would be to implement  PropertiesConfiguration from [Apache Commons Configuration](http://commons.apache.org/proper/commons-configuration//userguide/howto_properties.html#Properties_files)

e-fisher:
I'm not sure which IV is wrong, but they don't match: https://i.imgur.com/g5vVRew.png

pr0ves:
When catching the pokemon it printa the IVpercentage,  when transferring it primts the IV. 
Going to fix it right now.

neox387:
config file gets mixed up when saving the token

pr0ves:
Needs a way to try to login with username and password if the login with token fails.

Akkadius:
I was just thinking saving the token into config file would be nice, then you just went and implemented it

src/main/kotlin/ink/abb/pogo/scraper/Main.kt
```diff
@@ -93,7 +97,12 @@ fun main(args: Array<String>) {
     Log.normal("Logged in as $username with token ${auth.token.contents}")
 
     if (token.isBlank()) {
-        Log.normal("Set this token in your config to log in directly")
+        Log.normal("Setting this token in your config")
+        settings.setToken(auth.token.contents)
+        val out = FileOutputStream("config.properties")
+        settings.properties.store(out,null)
```
jabbink: Doesn't this fully rewrite the file and removes all comments?

src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
```diff
@@ -53,7 +53,8 @@ class CatchOneNearbyPokemon : Task {
 
                     if (result.status == CatchPokemonResponse.CatchStatus.CATCH_SUCCESS) {
                         ctx.pokemonStats.first.andIncrement
-                        var message = "Caught a ${catchablePokemon.pokemonId} with CP ${encounterResult.wildPokemon.pokemonData.cp} using $ball"
+                        val iv = (encounterResult.wildPokemon.pokemonData.individualAttack + encounterResult.wildPokemon.pokemonData.individualDefense + encounterResult.wildPokemon.pokemonData.individualStamina) * 100 / 45
```
jabbink: Please use encounterResult.wildPokemon.ivPercentage (or I am mistaken and that is in a different class, in which case extend the extension function in util/pokemon/Pokemon.kt)

src/main/kotlin/ink/abb/pogo/scraper/Main.kt
pr0ves: Yeah, it's because `properties` reformulates all of the properties and print on the file, resetting it first.
The only solution I found on the web is using PropertiesConfiguration from [Apache Commons Configuration](http://commons.apache.org/proper/commons-configuration//userguide/howto_properties.html#Properties_files) but I haven't tried yet.

src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
pr0ves: Yeah, it's another class, `pokemonData`. I spent ten minutes trying to find why it didn't let me use ivPercentage before figuring it out.

src/main/kotlin/ink/abb/pogo/scraper/Main.kt
jabbink: :/ Then I'd rather not have this in here because it will probably confuse people

pr0ves: I'll find a way to only modify the token

pr0ves:
Managed to fix it. Going to close this pr and open another to merge with develop.

PR #153, opened 2016-07-23
Change caught to collected

Pblolpb:
Phrasing of catching items sounds wrong. I think collecting is better.


pr0ves:
Well... it's "gotta catch 'em all" not "gotta collect 'em all".

Akkadius:
This PR is kind of ridiculous, not gonna lie

Pblolpb:
True, when you say it that way, I guess catch does make sense.  I just wasn't sure if it was it was supposed to be catch for both the pokemon and items.

pr0ves:
I'll close then.

Issue #34, opened 2016-07-21
Suggestion - Pokeball type by CP


PancakeZik:
I was wondering if you could add this, I'm at a high level and it would make sense to use pokeballs for low CP pokemon, great balls for medium and ultra for high CPs.

Could this be implemented?

Thank you very much!

falko100:
Already in there:
Program.cs:453-461

```
        // Use better balls for high CP pokemon
            if (masterBallsCount > 0 && pokemonCP >= 1000)
                return MiscEnums.Item.ITEM_MASTER_BALL;

            if (ultraBallsCount > 0 && pokemonCP >= 600)
                return MiscEnums.Item.ITEM_ULTRA_BALL;

            if (greatBallsCount > 0 && pokemonCP >= 350)
                return MiscEnums.Item.ITEM_GREAT_BALL;
```

PancakeZik:
I'm sorry, but should I just paste it anywhere on the main.kt?

falko100:
No need to do anything. It is already in the latest version on this repo. Just do a git pull and run it. Works fine :+1:

PancakeZik:
I'm sorry, but it's really not there.. Aren't you maybe thinking about some other bot?

falko100:
Sorry yea. C# bot :fearful:  Sorry :dancer:

CosmoRing:
For now i don't think it's possible because this bot is using the PokeGOAPI-Java and when you find a pokemon you get a EncounterResult and for now there is no way to get the pokemon data (CP and other stuff) from it.

PancakeZik:
I thought that might be the case, but then how about a first try with pokeball, then second try with great ball and third with ultra? Would that work?

CosmoRing:
I think PokeGOAPI now has all the needed code to implement this.
But i think there should be a change on how the bot is trying to catch pokemons. In this moment he tries to catch the pokemon using all the balls of one type. There should be a way to retry catching the pokemon in case the poke balls predefined are all used without success.

PancakeZik:
If somebody could just get the capture probability and make it decide which ball to use it would be great. I don't think changing the ball type would be that important.

jabbink:
Almost done with this feature.

I have the probabilities per ball, curve throw apparently does +10%. How much does a razz berry give?

That way the algo could be optimized to get a pokemon within 2 throws with 99% accuracy.

PR #168, opened 2016-07-24
added a threshold to use razzberries when capturing strong pokemons

ToMeRhh:
This 1800CP snorlax will never run away again!
(don't forget to --refresh-dependencies)


jabbink:
#34 

a better algorithm is incoming based on catch probability; not hardcoded CP levels

jabbink:
implemented in #174

PR #192, opened 2016-07-24
Show looted PokeStop name instead of ID

johnnyslt:
Before:

before

After:

after

I'm not much of a coder, so take it easy on me :)

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/LootOneNearbyPokestop.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/LootOneNearbyPokestop.kt
@@ -31,7 +31,8 @@ class LootOneNearbyPokestop(val sortedPokestops: List<Pokestop>, val lootTimeout
 
         if (nearbyPokestops.size > 0) {
             val closest = nearbyPokestops.first()
-            Log.normal("Looting nearby pokestop ${closest.id}")
+            val pokestopName = nearbyPokestops.first().details.name
+            Log.normal("Looting nearby pokestop \"${pokestopName}\"")
             ctx.api.setLocation(ctx.lat.get(), ctx.lng.get(), 0.0)
             val result = closest.loot()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/ink/abb/pogo/scraper/tasks/LootOneNearbyPokestop.kt
```diff
@@ -31,7 +31,8 @@ class LootOneNearbyPokestop(val sortedPokestops: List<Pokestop>, val lootTimeout
 
         if (nearbyPokestops.size > 0) {
             val closest = nearbyPokestops.first()
-            Log.normal("Looting nearby pokestop ${closest.id}")
+            val pokestopName = nearbyPokestops.first().details.name
```
CosmoRing: @jabbink said that this adds one HTTPS request every looted log in #36 

You really want to add this added, maybe it's a good idea to add an option on the config file and a warning saying it will increase https requests.

jabbink: Maybe rename the settings.shouldDisplayWalkingToNearestUnused and use the same flag for both of these.

johnnyslt: Sadly, I don't know how to do that :( 
This was the best thing I could do with my knowledge.

If anyone could take this ideea and make like @jabbink said, that would be great

CosmoRing: I can do that, but will have to see how do i add to a pull request that is not mine :)

johnnyslt: Make a new one, and I'll close mine :D

Issue #179, opened 2016-07-24
option to end bot at Encounter failed with result: POKEMON_INVENTORY_FULL 


cleftfairy:
When this happens, the bot just keeps merrily along which generates 5 second apart or so attempts to encounter the same pokemon - seems like it could trigger an anti-ban. I'd prefer the bot exit at that point or have some way to dump pokemon. I reached level 22 and caught too many pokemon over cp 400...

ToMeRhh:
you're right that there's no reason to continue encountering pokemons, but I think the right thing to do is to move to 'walk mode'.

obviously the solution for full pokebank is simply to raise the CP/IV threshold :)

PR #205, opened 2016-07-24
Feature walkOnly when Pokemon Inventory Full

krystiancharubin:
**Implemented feature:** #179

**Changes made:**
changes Settings.walkOnly to var so we can change value
sets Settings.walkOnly to true when encounterResponse.Status is POKEMON_INVENTORY_FULL


jabbink:
Inventory can sometimes be full for a short time (when for some reason the server fails and doesn't throw away your stuff) so there should be a method to link this to the dropUselessItems thing (eg. when that one is off, and then you encounter that it's full you go to walk only mode).

Also personally I wouldn't want this in my own bots config, as even when your inventory is full, you get +50XP per activated pokestop.

PR #206, opened 2016-07-24
Sorted Pokemon List

CosmoRing:
**Changes made:**

* Sorted Pokemon List displayed after login

Just a quick sort to pokemon list shown after login.


jabbink:
Implemented (slightly different, now sorted by name, and within name by descending IV)

PR #208, opened 2016-07-24
Added settings toggle for sorting by IV instead of CP

Forty-Tw0:
Even with transfer_iv_threshold and transfer_cp_threshold set to -1, transfer will still favour the pokemon with the highest CP even if it has an IV of 1%. I propose the setting sort_by_iv to allow users to keep the highest IV pokemon regardless of CP.

### DIFF
--- a/config.properties.template
+++ b/config.properties.template
@@ -56,6 +56,9 @@ autotransfer=false
 # minimum amount of pokemon type to keep
 keep_pokemon_amount=1
 
+# Sort by IV first instead of CP
+sort_by_iv=false
+
 # Minimum IV percentage to keep a pokemon (to ignore IV: use -1)
 # between 0 and 100, suggested 80
 transfer_iv_threshold=80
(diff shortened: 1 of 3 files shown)


pashorizer:
I don't think this is a good idea since sorting by IV would release high CP Pokemon with low IV.  I was messaging earlier with a mate about this topic and he send me this

>   "Also, technically, because the max CP multiplier is a fraction (79.x%), the real stats of pokemon will only differ by, at max level, ~12, i.e., a level 40 Dragonite with 0 IVs will only have 12 less attack, defense, and HP than does a Dragonite with max level and IV's."

I am still digging into this topic, so we might need a better way of determining when to release a pokemon or rate one over the other one.

Forty-Tw0:
For the long run, higher IVs are better because you can always use stardust to up the CP. Two pokemon with equal CP may not be equal in stats. IVs range from 0-15 so a 1000CP pokemon with 0% IV quality will have base stats, while a 1000CP pokemon with 100% IV quality will have base stats +15. You can always powerup a lower CP pokemon with stardust, but you can not change IVs. Since this bot results in tens of thousands of stardust (and many candy), I would prefer perfect IV pokemon over anything else. The proposed change is a toggle which defaults to false, I suspect only 'advanced' trainers would care about the minuscule IV boosts anyway.

PR #218, opened 2016-07-24
Moved the shouldAutoTransfer check before starting the thread

pr0ves:
Same as #147

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
@@ -126,7 +126,7 @@ class Settings(val properties: Properties) {
 
         file.lines().forEach {
             if (it != null && it.startsWith("token")) {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
```diff
@@ -112,4 +115,30 @@ class Settings(val properties: Properties) {
...
+        properties.setProperty("token", value)
+    }
+
+    fun writeToken(propertyFile: String) {
+        val file = BufferedReader(FileReader(propertyFile))
+        var propertiesText = String()
+
+        file.lines().forEach {
+            if (it != null && it.startsWith("token")) {
+                propertiesText += "token = ${this.properties.getProperty("token")}\n"
```
jabbink: Does this work with the spaces before and after the `=`?

pr0ves: Yeah, it does.

PR #223, opened 2016-07-24
Spring boot

apottere:


apottere:
Not sure why it couldn't find the latest version of the api - I can pull that revision just fine locally.

PR #262, opened 2016-07-25
Updated Login

Darakath:
Stores refresh tokens in bot settings files and uses it to login.


jabbink:
Already fixed by reverting

PR #302, opened 2016-07-25
Brighter colors for better readability

Darakath:


jabbink:
Can you add comparison screenshots?

CosmoRing:
image

MASSV:
Worth noting too is colors seem to be effected by settings for terminal/console. Colors for me are not the same on Windows CMD.

classicalDonkey:
Can we have different colours for different things?  I like that something are highlighted now, but we like have a different colour for Hatching, Looting, Caught etc

PR #308, opened 2016-07-25
Fix invalid token written in config.properties

Demelza:
Has been tested locally and works perfectly.


jabbink:
/facepalm

Thank you.

PR #326, opened 2016-07-25
Remove internal pokestop blacklist in favor of the protobuff blacklist.

apottere:
Why were we using a custom lootTimeout instead of the timeout provided by the server?  With the custom timeout, the bot would try to loot all the pokestops it had the run before (if restarted within 5 mins from the same location).  Using the server timeouts, it works just fine.

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/LootOneNearbyPokestop.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/LootOneNearbyPokestop.kt
@@ -21,13 +21,13 @@ import ink.abb.pogo.scraper.util.map.canLoot
 import java.util.*
 import java.util.concurrent.TimeUnit
 
-class LootOneNearbyPokestop(val sortedPokestops: List<Pokestop>, val lootTimeouts: HashMap<String, Long>) : Task {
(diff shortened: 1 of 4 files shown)


jabbink:
The list was introduced because the server very often reports `0` as timestamp until activation.
Even after a full map objects refresh.

https://github.com/jabbink/PokemonGoBot/issues?utf8=%E2%9C%93&q=cooldown%20is%3Aissue

jabbink:
(and only directly after looting the timestamp would be correct)

apottere:
Well it should at least respect the server timeout if it exists - I can do that.

jabbink:
Yeah, that's why it puts `lootTimeouts.put(closest.id, closest.cooldownCompleteTimestampMs)` the correct timestamp there and only when it's not sure at all and gets the cooldown error, it sets a hardcoded 5 minutes.

On another note: many parts of the code (also of the Java API) rely on the system time. Maybe make it system time independent and somehow include a synced, correct timer.

apottere:
Right, but it only puts the timeout in there if it's successful.  That doesn't help on a fresh run, because this instance of the bot hasn't tried to loot it yet.

PR #327, opened 2016-07-25
Respect both our pokestop timeout *and* the server timeout.

apottere:
If we haven't visited a pokestop yet, assume the server knows the timer.

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/util/map/Pokestop.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/util/map/Pokestop.kt
@@ -11,7 +11,6 @@ package ink.abb.pogo.scraper.util.map
 import com.pokegoapi.api.map.fort.Pokestop
 
 fun Pokestop.canLoot(ignoreDistance: Boolean = false, lootTimeouts: Map<String, Long>): Boolean {
-    val now = System.currentTimeMillis()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/ink/abb/pogo/scraper/util/map/Pokestop.kt
```diff
@@ -11,6 +11,7 @@ package ink.abb.pogo.scraper.util.map
 import com.pokegoapi.api.map.fort.Pokestop
 
 fun Pokestop.canLoot(ignoreDistance: Boolean = false, lootTimeouts: Map<String, Long>): Boolean {
-    val canLoot = lootTimeouts.getOrElse(id, { 0 }) < System.currentTimeMillis()
+    val now = System.currentTimeMillis()
+    val canLoot = lootTimeouts.getOrElse(id, { 0 }) < now && cooldownCompleteTimestampMs < now
```
jabbink: Wouldn't `val canLoot = lootTimeouts.getOrElse(id, { cooldownCompleteTimestampMs }) < now` work as well? As far as I know this `Map` is not filled with 0's or values it knows are incorrect.

apottere: Yup, I wrote this before the comment on the other PR - I didn't realize it stored blacklist timers for a successful spin.

apottere:
Feel free to squash when you merge.

apottere:
Also, travis is really hurting.

PR #330, opened 2016-07-25
gracefully try to login multiple times

jabbink:
**Changes made:**

* Try to get auth ticket with stored token
* in case of failure: remove token and retry again with username/password
* if that succeeds: store new token
* else: die

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Main.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Main.kt
@@ -9,7 +9,9 @@
 package ink.abb.pogo.scraper
 
 import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass
+import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope.AuthInfo
 import com.pokegoapi.api.PokemonGo
+import com.pokegoapi.api.pokemon.Pokemon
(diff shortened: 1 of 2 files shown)


sphinxid:
this one is really important, at least user don't need to remove the token manually from the config file.
you should merge this if there isn't any conflict.

jabbink:
I heard from some testers that this breaks the Google login even more :/

Unfortunately nobody that dares to use his/her Google account with this tried to debug the code.

sphinxid:
Ah, I see, unfortunately, I only have PTC account for testing purpose. Maybe we could refactor the code a bit to cater graceful login on PTC account only (for now).

pr0ves:
I found that the problem happens [here ](https://github.com/jabbink/PokemonGoBot/blob/2bcda4a60dfb93ea41080d9ecd7f1c1bec8f87c0/src/main/kotlin/ink/abb/pogo/scraper/Main.kt#L62-L64)
`if (auth.provider.equals("google")) {    
        displayToken = GoogleLoginSecrets.refresh_token    
    }`

the `GoogleLoginSecrets.refresh_token` sets the token to null

jabbink:
Not exactly, it is initialized with `null` and `refreshToken(String ...)` doesn't set it.

jabbink:
Could you try again?

jabbink:
I tested it for PTC.

@pr0ves for Google
image

Merging.

PR #331, opened 2016-07-25
Speed up travis builds

Machtergreifung:
**Changes made:**

* Removed sudo, so travis will start up a VM faster. See https://docs.travis-ci.com/user/ci-environment/


jabbink:
Nice catch, didn't even notice that I copypasted that in.

PR #334, opened 2016-07-25
Added randomness to next pokestop selection

citruz:
**Changes made:**

* Made pokestop selection non deterministic by choosing a random pokestop out of the nearest five while still taking the distance into account

Background: I noticed that my bot was sometimes stuck in a loop and always visited the same 10-20 pokestops (after they were out of cooldown). This resulted in fewer pokemons caught since they did not respawn in the meantime.
With this change the bot will not always select the nearest pokestop to go to. Instead the next pokestop is chosen randomly from the 5 nearest ones. To prevent cases where the bot selects a pokestop which is far away, the probability is not the same for each pokestop. It is proportional to the distance meaning that a pokestop which is very close will have a higher probabilty to be chosen (but it is not always chosen).
This will introduce some randomness but still maximize the XP gained.


Emuuung:
This feels nice, but maybe add it as an option in `config.properties.template` as a boolean.

jabbink:
Oh lol, just too late as I just merged it.

Emuuung:
Ahh yeah, I noticed lol

PR #345, opened 2016-07-25
Allow for graceful bot shutdown.

apottere:
Currently, when killing the bot (SIGINT), the bot dies immediately, regardless of whether it was executing an API call.  This change allows the bot loops to finish before the bot is killed.

Consequently, there's another side effect - the bot loops are now synchronous.  This is also an improvement, as you don't ever want your bot loop running twice at the same time.

With this change, the loop allocate 5 seconds for the bot loop, and depending on how long the actions take will only "sleep" for the remaining time after the tasks are done.  I wasn't sure if that would be preferable to always waiting 5 seconds after tasks were complete.

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
@@ -12,17 +12,28 @@ import com.google.common.util.concurrent.AtomicDouble
 import com.pokegoapi.api.PokemonGo
 import com.pokegoapi.api.player.PlayerProfile
 import com.pokegoapi.api.pokemon.Pokemon
-import ink.abb.pogo.scraper.tasks.*
+import ink.abb.pogo.scraper.tasks.CatchOneNearbyPokemon
+import ink.abb.pogo.scraper.tasks.DropUselessItems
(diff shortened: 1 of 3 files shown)


apottere:
In case the code is a little confusing, the bot "sleeps" in the while loop using a countdown latch.  This allows `latch.countDown()` to simultaneously "break" the loops from their sleep and make the while loop exit.  The threads use the phaser to let the bot know they're done, and the stop method returns.

apottere:
Most importantly, a shutdownHook is registered in the main method that calls `bot.stop()`.

jabbink:
:+1: 
>depending on how long the actions take will only "sleep" for the remaining time after the tasks are done
Sounds good. What happens when the loop takes longer than 5 seconds? If I understand correctly it won't start that loop again then? If so, great!

apottere:
If the loop takes longer than 5 seconds, it starts a new loop immediately after.

apottere:
Since usually the only thing making the loop take a while is transferring pokemon after having autotransfer off, this shouldn't be that suspicious.

jabbink:
Will test this in a minute (not at pc with compiler right now) and merge when it works.

PR #98, opened 2016-07-23
Added auto-evolving.

Javapt:
Required candy is read from new file 'pokemon-candy.csv'
Added new configuration option which accepts a list of a pokemon to auto evolve.

PR #347, opened 2016-07-25
Auto Evolve

xenxes:
Bringing #98 back to develop.  Please test and modify accordingly.

Additional recommended feature --

IF PokeBank > # (or FULL) && candies > #
  USE Lucky Egg
  EVOLVE for 30 Mins


src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
```diff
@@ -69,6 +69,7 @@ class Bot(val api: PokemonGo, val settings: Settings) {
         val catch = CatchOneNearbyPokemon()
         val release = ReleasePokemon()
         val hatchEggs = HatchEggs()
+        val evolve = EvolvePoekmon()
```
CosmoRing: Poekmon :+1:

src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
```diff
@@ -94,11 +95,12 @@ class Bot(val api: PokemonGo, val settings: Settings) {
 
                 task(process)
                 task(hatchEggs)
+                task(evolve)
```
xenxes: How do you check for a null string if auto_evolve in config is empty?

Via != null or .length?  Strings can't be null in kotlin right?

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
```diff
@@ -93,6 +94,17 @@ class Settings(val properties: Properties) {
         listOf()
     }
 
+    val autoEvolve = getPropertyIfSet("list of pokemon you want to evolve when able to", "auto_evolve", "CATERPIE,PIDGEY,WEEDLE", String::toString).split(",")
+
+    //This method only exists because I wasn't sure if this data was stored somewhere else. If it is then this can be removed.
+    private fun getCandyByPokemon(): Map<Int, Int> {
+            val lines = File("pokemon-candy.csv").readLines()
```
apottere: Could this be in `src/main/resources` and loaded from the classpath?

xenxes: I think it is in PokemonMeta.java if someone wants to mod and remove .csv --

	public PokemonMeta(	int baseStam,
						double baseCaptureRate,
						int candiesToEvolve,

src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
jabbink: `String`s can't be null. But if you most likely have a `String?`

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
MASSV: "IF PokeBank > # (or FULL)
USE Lucky Egg
EVOLVE for 30 Mins"

as far as lucky egg I think it would be better to have bot count number of possible evolutions from listing in config aka


+# list of pokemon you want to auto evolve
 +auto_evolve=PIDGEY,weedle,caterpie
+# Minimum Number of evolutions possible before Auto, Leave at -1 or blank to evolve always.
 +auto_evolvemin=50
+# Use Lucky Egg before Evolve cycle
 +auto_evolveUseEgg=true

so that would wait for 50 possible evolutions from that list, then use lucky egg.

+# list of pokemon you want to auto evolve
 +auto_evolve=PIDGEY,weedle,caterpie
+# Minimum Number of evolutions possible before Auto, Leave at -1 or blank to evolve always.
 +auto_evolvemin=-1
+# Use Lucky Egg before Evolve cycle 
 +auto_evolveUseEgg=false

Would evolve every pidgey etc after a catch. no egg

xenxes: That's a good idea, I just ran this and it evolved 87 Pidgies in less than a minute (build in delays?).  Perhaps best to leave lucky egg usage on manual, they're rare to come by without spending.

Akkadius:
Auto-evolve in release would be amazing

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
Killberty: @xenxes i suggest letting the user decide how mnay evolutions/minute he wants (-1 of unlimited)

src/main/kotlin/ink/abb/pogo/scraper/tasks/EvolvePokemon.kt
```diff
@@ -0,0 +1,49 @@
...
+        val autoEvolve = settings.autoEvolve
+        val canEvolve = groupedPokemon.filter {
+            val candyNeeded = settings.candyRequiredByPokemon[it.key.number]
+            candyNeeded != null && candyNeeded > 0 && autoEvolve.contains(it.key.name) && it.value.first().candy >= candyNeeded
+        }
+        if (canEvolve.isEmpty()) {
+            return
+        }
+        canEvolve.forEach {
+            val sorted = it.value.sortedByDescending { it.cp }
```
Forty-Tw0: Can we use the sortByIV setting here like in #208.

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
Akkadius: @xenxes : do you have a link to a compiled jar of this to check out? I can't compile because windows is stupid and I get too long of a file path when compiling dependencies.

jabbink: >windows is stupid and I get too long of a file path 

Clone the repo high on your SSD/HDD, eg. in `C:\`

Agronis:
May I suggest with a future adaptation to this that a few things are taken into account.

If the account is missing X from Pokedex, make this a priority in evolution chain.

IE:  I have 200 Poliwag candy.. but I'm missing a Polywrath from my dex.

Currently, I'd be left with 8 Poliwhirl.  Where as I need a Polywrath.


# Evolve highest IV/CP Pokemon into missing Pokedex entry.
# Uses closer evolution if IV/CP is equal to or greater than lower evolution tier.

pokedex_priority=true/false


Now, I'd receive one Poliwrath and two Poliwag (presuming Wrath came from Tier1 evolution).

jabbink:
Afaik the required candies per evolution are now available in the api (PokemonMetaRegistry or sth like that)

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
Akkadius: I ended up doing that, which worked.

KyleBoyer:
Can you make an option in the config to say 'ANY', thus is would evolve any Pokemon that have enough candies.

src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
```diff
@@ -85,6 +86,7 @@ class Bot(val api: PokemonGo, val settings: Settings) {
         fixedRateTimer("BotLoop", false, 0, 5000, action = {
             thread(block = {
                 task(keepalive)
+                task(evolve)
```
bolte-17: I like this moved down after task(catch). Otherwise those Pidgeys get released before we can evolve them for the XP.

woutermb:
@jabbink is this branch going to be included in the next release? would love to see something i can use all the excess candy for, and evolving via API is way faster than manual! :)

src/main/kotlin/ink/abb/pogo/scraper/tasks/EvolvePokemon.kt
megatixos: Yea just change this: val sorted = it.value.sortedByDescending { it.cp } with this: 
            val sorted = if (settings.sortByIV) {
                it.value.sortedByDescending { it.getIv() }
            } else {
                it.value.sortedByDescending { it.cp }
            }
I dont know how to make a pull request so i only can write it here sorry!

xenxes:
closing in favor of https://github.com/jabbink/PokemonGoBot/pull/444 (no csv)

Issue #338, opened 2016-07-25
Use \r\n linebreaks in Settings writer


jabbink:
**Description:**
Windows: Notepad.exe does not render `\n` as new line.

**Steps to reproduce:**

1. start bot on Windows
2. token gets set
3. Open config in notepad
4. one long line; pressing Ctrl+S breaks the entire file

**Expected behavior:**
show newlines `\r\n`

**Actual behavior:**
single long line; `\n`

**Version:**
`develop`

PR #352, opened 2016-07-25
added more iv details

pashorizer:
Looks like this
stats1
stats2


jabbink:
The line break thing shouldn't be necessary. #338 is only about the settings writer.

PR #363, opened 2016-07-26
Always Throw Curveballs Option

Enochen:


Lenovotro:
Is there a downside to always throwing curveballs? Haha

Enochen:
It defaults to not throwing unless needed right now that's why lol

jabbink:
>Is there a downside to always throwing curveballs?

Makes it even easier for Niantic to detect the use of bots.

config.properties.template
```diff
@@ -69,7 +69,10 @@ autotransfer=true
 keep_pokemon_amount=1
 
 # Sort by IV first instead of CP
-sort_by_iv=false
+sort_by_iv=true
+
+# Always throw a Curveball
+always_curve=true
```
jabbink: Set default to false please.

config.properties.template
```diff
@@ -69,7 +69,10 @@ autotransfer=true
 keep_pokemon_amount=1
 
 # Sort by IV first instead of CP
-sort_by_iv=false
+sort_by_iv=true
```
jabbink: Don't change the defaults of other variables

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
```diff
@@ -77,6 +77,8 @@ class Settings(val properties: Properties) {
     val shouldAutoFillIncubatores = getPropertyIfSet("Auto Fill Incubatores", "auto_fill_incubator", true, String::toBoolean)
 
     val sortByIV = getPropertyIfSet("Sort by IV first instead of CP", "sort_by_iv", false, String::toBoolean)
+    
+    val alwaysCurve = getPropertyIfSet("Always do a Curveball", "always_curve", true, String::toBoolean)
```
jabbink: also here, `false`

jabbink:
as this has merge conflicts I will put this in #399

Issue #357, opened 2016-07-25
[SUGGESTION] Changes to Console Output


Killberty:
**Description:**

- Make status outputs like these blue, because they are really hard to find, escpecially if some exceptions get printed:
  - Profile update: 1081915 XP on LVL 26; 181915/200000 (90.96%) to LVL 27
  - XP gain: 3760 XP; Pokemon caught/transferred: 13/11; Items caught/dropped: 0/0; Pokebank 171/250; Stardust 342688
- Make rare events orange
  - Egg hatching and and putting eggs in incubators (thx @Agronis )
  - Level ups maybe
- Allow filtering
  - Every color has a different meaning (like red for error, green for sucess, blue for status eventually etc.) so maybe allow filtering Output by color
  - Also allow filtering for exceptions (i know it has been requested already, but it fits here)

**Version:**
0.4.2

pascalpfeil:
_

Agronis:
I'll toss in a specific request for Egg updates.

WebmasterNikos:
Consider making the username part of the profile update, would make it easier to tell which console is which when running multiple bots.

jabbink:
Implemented I think.

Issue #353, opened 2016-07-25
[Suggestion] Color Code Profile Updates


Scoboose:
Can you please color code profile updates?  Would be great to at a glance see the last update with out searching for it.

Scoboose:
Yes #357 Is what I am asking.

omar972:
If you compile from source, you would be satisfied.

jabbink:
Implemented.

Scoboose:
Thanks!

PR #367, opened 2016-07-26
[Log Console] Facilitate and perhaps answer #357 and #353.

joligario:
Define remaining standard ANSI colors and make white bold.
Set profile update to use white.
Set egg events to use cyan.


Emuuung:
Looks nice but a few logs still look "old white" like
```
Destination reached.
Walking to (xx.xxxxx, xx.xxxxxx) in xxx.xxxxxxxx steps.
Looting nearby pokestop xxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
And after catching a pokemon, log details of IV's and rewards for catching have a "dark green".

joligario:
Yes, that is because it is still set as 'normal' and not yet assigned a color.

Issue #391, opened 2016-07-26
[Suggestion] Different desired catch rate based on CP/IV


PancakeZik:
I think we have all the data needed to implement this.
How about a catch rate for when the pokemon doesn't fit your "transfer criteria" and another one for when it does?
I think that if I see a crappy pidgey now and the bot has to waste an ultra ball + razz berry to catch it, it could be saving that ultra ball for a more important pokemon later on.

jabbink:
Closing #375 in favor of this one

Mewks:
Great idea, force the use of Pokeballs only on Pokemon we are going to transfer. If we arent going to transfer them we use the probability criteria as normal?

PR #399, opened 2016-07-26
Introduce lower catch probability for unwanted Pokemon

jabbink:
**Fixed issue:** #391
**Changes made:**

* Lower the default desired catch probability
* Introduce second desired catch probability for unwanted pokemon

### DIFF
--- a/config.properties.template
+++ b/config.properties.template
@@ -11,7 +11,8 @@ latitude=
 longitude=
 
 # desired probability to catch Pokemon with 1 throw
-desired_catch_probability=0.5
+desired_catch_probability=0.4
+
 # desired probability to catch unwanted Pokemon (obligatory_transfer; low IV; low CP)
 # set to -1 to not attempt catching at all
 desired_catch_probability_unwanted=0.0
@@ -76,6 +77,9 @@ sort_by_iv=false
 
 # Always throw curveballs
 always_curve=false
+
+# Never use berries
+never_use_berries=true
 
 # Minimum IV percentage to keep a pokemon (to ignore IV: use -1)
 # between 0 and 100, suggested 80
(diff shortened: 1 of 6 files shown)


jabbink:
Should add this after the encounter is complete, but need to add more to prevent the bot getting stuck every 5 seconds again on the same Pokemon when this occurs.

```kotlin
                if (desiredCatchProbability == -1.0) {
                    Log.normal("CP/IV of encountered pokemon ${catchablePokemon.pokemonId} turns out to be too low; skipping")
                    return
                }
```

PancakeZik:
I'm getting reason: null for pokemon on the unwanted list
26 Jul 12:32:37: Going to transfer DROWZEE with CP 389 and IV 71%; reason: null

jabbink:
Fixed.

jabbink:
https://github.com/jabbink/PokemonGoBot/pull/399#issuecomment-235239648 implemented.

config.properties.template
```diff
@@ -77,6 +78,9 @@ sort_by_iv=false
 # Always throw curveballs
 always_curve=false
 
+# Never use berries
+never_use_berries=true
```
alawing: Did this will blocked all berry using? My suggestion is berries not using with redball together, change to blue ball first, if probability if also low, try berry then.

jabbink: That is already the default behavior for days.

It will first try ever higher balls. Then add curve. Then add a berry.

alawing: I don't think so, because yellowball is harder to get than berry. if sometime I prefer using blue+berry than yellowball. But the logic you give is try yellowball first. So my conclution is
red<blue<blue+berry<yellow<yellow+berry
yours is
red<blue<yellow<yellow+berry

jabbink: There is no hardcoded ball order. It calculates it by the catch probabilities it receives directly from the server.

PR #405, opened 2016-07-26
Migrate to new login system

jabbink:
**Changes made:**

* Migrate the login mechanism to the new API way to do this
* Removed call to now deprecated function


jabbink:
@pr0ves can you test Google Login?

CosmoRing:
@jabbink Just tested Google Login in the new Login System and it works.

PR #406, opened 2016-07-26
Curve all the time and save berry

alawing:
Changed Catch pokemon logic
*Changed to using curve ball all time
*Save berries if using basic pokeball (because current logic consume too many berries)


jabbink:
Why remove so much functionality? Also implemented in a configurable way in #399

alawing:
I just makes force curveball, and also add one condition if using redball, dont use berry

jabbink:
Again: see #399 

adds all that as options

PR #418, opened 2016-07-26
Prevent redball+berry

alawing:
redball+berry will consume too much berries, so try use blueball first than berry. And still could use berry if highestball is redball.


smileytechguy:
I get berries all the time. How practical is this?

PR #419, opened 2016-07-26
Adds Configuration for randomness to next pokestop selection

CosmoRing:
**Expands issue:** #334

**Changes made:**

* Added `randomNextPokestop` to config file

Now you can adjust the randomness of your next pokestop selection (taking into account the nearest ones). If you don't want randomness just put 1 in the field.

### DIFF
--- a/config.properties.template
+++ b/config.properties.template
@@ -17,6 +17,10 @@ desired_catch_probability=0.4
 # set to -1 to not attempt catching at all
 desired_catch_probability_unwanted=0.0
 
+# number of nearest unused pokestops to randomly select
+# the next pokestop from (1/infinit)
+randomNextPokestop=1
+
 ## Walking
 # Meters per second
 speed=2.778
--- a/config.properties.template
+++ b/config.properties.template
@@ -18,8 +18,8 @@ desired_catch_probability=0.4
 desired_catch_probability_unwanted=0.0
 
 # number of nearest unused pokestops to randomly select
-# the next pokestop from (1/infinit)
-randomNextPokestop=1
+# the next pokestop from (1/infinite)
+random_next_pokestop_selection=5
 
 ## Walking
 # Meters per second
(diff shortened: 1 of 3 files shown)


config.properties.template
```diff
@@ -17,6 +17,10 @@ desired_catch_probability=0.4
 # set to -1 to not attempt catching at all
 desired_catch_probability_unwanted=0.0
 
+# number of nearest unused pokestops to randomly select
+# the next pokestop from (1/infinit)
```
jabbink: *infinite

config.properties.template
```diff
@@ -17,6 +17,10 @@ desired_catch_probability=0.4
...
+randomNextPokestop=1
```
jabbink: Can you rename this so this is named as others, eg `random_next_pokestop_selection`.

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
```diff
@@ -63,6 +63,8 @@ class Settings(val properties: Properties) {
         )
     }
 
+    val randomNextPokestop = getPropertyIfSet("Number of pokestops to select next", "randomNextPokestop", 5, String::toInt)
```
jabbink: Don't forget to edit that name in parameter 2 here as well

KyleBoyer:
So is -1 infinite, or is "infinite"?

CosmoRing:
It means 1, 2, 3, ... + infinite.

It starts on 1 and goes up.

KyleBoyer:
@jabbink @CosmoRing I think then it should be 2,147,483,647 as the maximum, as anything over that is not a Java Integer.

jabbink:
Oh wow, I didn't expect anyone to actually think they should put the literal word "infinite" in there. That would be an utter waste of time and XP (and yes, it would crash, but even if it was implemented to work).

PR #434, opened 2016-07-26
Show reason for unwanted probability when chosen

jabbink:
**Changes made:**

* Help people understand why a certain probability was selected


KyleBoyer:
26 Jul 14:54:11: Using desired_catch_probability_unwanted because Obligatory transfer 
26 Jul 14:54:11: Using ITEM_GREAT_BALL; Using curve; achieved catch probability: 0.4680462, desired: 0.0 

I still want to catch the pokemon in my obligatory transfer list, just not keep them - I want the candies. I keep the highest evolution of pokemon, and want the candies to evolve/upgrade those pokemon further.

jabbink:
? In this case it did catch the Pokemon, didn't it (you excluded the actual line that would tell anyone...)?

KyleBoyer:
But the main point is: 

Using desired_catch_probability_unwanted because Obligatory transfer 

Because it is a wanted pokemon for its candies, just not for it taking up space in my inventory.

Issue #457, opened 2016-07-26
[Feature Request] Adjustable timer for profile update.


sodanakin:
**Adjustable time for Profile Update Message.**

I'd like to be able to adjust the time between displaying each profile update.
I am currently editing this manually each time at a new release.

Thnx :)

Edit : Giving it a shot myself to add it in.. Will see if it works if so ill make a pull request :)
Edit2 : Got it working and separated it from the hatcheggs timer :)
Edit3 : Made a pull request :D

sodanakin:
Request filled by merge.

PR #460, opened 2016-07-27
Adjustable profile update.

sodanakin:
**Fixed issue:** [Reference the issue number here, or remove if not a fix]

#457 (my own request lol)

**Changes made:**

I split the Hatcheggs and profile loop into two loops whereby I added control over the timer of the profile loop. Which you can now adjust in the config.
Updated the settings.kt file accordingly.

### DIFF
--- a/config.properties.template
+++ b/config.properties.template
@@ -85,6 +85,9 @@ always_curve=false
 # Never use berries
 never_use_berries=true
 
+# Set Timer for profile,Xp,lvl update. (Default: 60)
+set_timer=60
+
 # Minimum IV percentage to keep a pokemon (to ignore IV: use -1)
 # between 0 and 100, suggested 80
 transfer_iv_threshold=80
--- a/config.properties.template
+++ b/config.properties.template
@@ -85,8 +85,8 @@ always_curve=false
 # Never use berries
 never_use_berries=true
 
-# Set Timer for profile,Xp,lvl update. (Default: 60)
-set_timer=60
+# Set Profile Update Timer (Default: 60)
+profile_update_timer=60
 
 # Minimum IV percentage to keep a pokemon (to ignore IV: use -1)
 # between 0 and 100, suggested 80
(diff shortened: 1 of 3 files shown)


config.properties.template
```diff
@@ -85,6 +85,9 @@ always_curve=false
 # Never use berries
 never_use_berries=true
 
+# Set Timer for profile,Xp,lvl update. (Default: 60)
+set_timer=60
```
jabbink: Please rename to profile_update_timer

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
```diff
@@ -18,6 +18,8 @@ import java.util.*
 
 class Settings(val properties: Properties) {
 
+	val profileUpdateTimer = getPropertyIfSet("Set Timer", "set_timer", 60, String::toLong)
```
jabbink: if you do the fix from above, edit it here as well

src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
```diff
@@ -98,8 +98,11 @@ class Bot(val api: PokemonGo, val settings: Settings) {
         runningLatch = CountDownLatch(1)
         phaser = Phaser(1)
 
-        runLoop(TimeUnit.SECONDS.toMillis(60), "ProfileLoop") {
+        runLoop(TimeUnit.SECONDS.toMillis(settings.profileUpdateTimer), "ProfileLoop") {
             task(profile)
+        }
+		
+	runLoop(TimeUnit.SECONDS.toMillis(60), "HatchEggsLoop") {
             task(hatchEggs)
```
jabbink: not sure if hatchEggs relies on the profile update (it does need a fresh inventory). If that's the case, please remove this runLoop and put the eggs behind the task(profile) again

sodanakin:
Can't test atm PTC is down I will edit it in a bit when servers aren't puking rainbows :")
Also about the HatchEggs I had the bot running with this implemented since I added it without issues. So do I need to put it back in the same loop or not? Just want to be sure what to edit before making a new pull request. @jabbink

Issue #450, opened 2016-07-26
Create nice interface for config creation


jabbink:
**Description:**
The config is getting out of hand. A nice (web) interface to click a config together would be nice. This interface should group options logically together, visually show which option supersedes another (`autotransfer=false` disables **many** other parameters completely).

pr0ves:
Maybe as a page to @shilch's webui

shilch:
I thought of adding a window 'Settings' to the GUI. But there would only be options that can be changed during runtime.

pr0ves:
Maybe a Settings page for live changes, and a ConfigEditor one to change the config file

jabbink:
I actually thought about a simple HTML + JavaScript page hosted on GitHub pages that can read the properties file and create a new one that can be copypasted back.

shilch:
Nice idea, I will try :D

chirgjin:
I made one [Here](http://pkmn-go.in/web-config.php)
Be sure to post ur suggestions ;-;

langerhans:
❗ **ATTENTION** ❗ 
Keep in mind when you use above website you should probably not put your credentials in and instead do that manually. Not saying this site specifically is malicious, just that you should in general be careful. 
❗ **ATTENTION** ❗

Taclaa:
You could also have a map on this page so you can select the bot lat/long.

shilch:
I started creating one 3 weeks ago but due to lack of time I was not able to finish it. Maybe someone wants to complete the creation tool? https://gist.github.com/shilch/cf75bb509e3fc39a52fc22e9886a17ca

shilch:
I used a map there @Taclaa .
![Screenshot](https://cloud.githubusercontent.com/assets/11890358/17727969/f856655c-645c-11e6-8431-d157e48a9110.png)

Taclaa:
Nice! @shilch

shilch:
Feel free to finish it if you have time.

DiogoRNeves:
I can try to give it a go tomorrow (in about 9/10 hours), I'll join you guys on Discord and let you know.

DiogoRNeves:
Update: I changed @shilch 's work to fetch the most recent template from GitHub and replace the options in the page. Added some options, just have to add the rest of them.

[...]

PR #233, opened 2016-07-24
Added Web-GUI

shilch:
See #197 for To-dos
Added a Web-GUI (beta, I will add much more)
- Webserver is running on port 8000
- Socketserver is running on port 8001

bot

PR #483, opened 2016-07-27
Config GUI

ZingBallyhoo:
A configuration GUI built in Python and Tkinter

[Image](http://i.imgur.com/wS6y7Dc.png)
Note: 
* Using https://github.com/shilch/PokemonGoBot (#233)

This possibly solves #450?

**Features:**
* Read options from config.properties.template
* Can create config.properties from scratch
* Reads existing config.properties
* Reset to defaults
* Save to config.properties (Otherwise there is no point in this)
* Tries to auto detect true/false options
* The text for labels (next to the entries) are created on the fly, and are not hard coded

**Things I would like to do:**
* When the template is updated add the new entries to config.properties without a reset
* Code cleanup
* Make it look nicer
* Grouping of entries

**Something I would like to do but I'm not sure if I have the skills:**
* Google maps GUI for picking Long and Lat
* Possibly using https://github.com/simondlevy/GooMPy ?

**Issues:**
* Coded in Python, and the bot is in Java, but I don't think that is too big of a deal
* Requires Python 3, I might make it compatible with 2 if I get time

### DIFF
--- a/configTool.pyw
+++ b/configTool.pyw
@@ -13,6 +13,7 @@ word = []
 val = []
 name = []
 entries = []
+saved_lines = []
 done = False
 tf = ["true", "false"]
 cRow = 0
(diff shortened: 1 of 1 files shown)


ZingBallyhoo:
Groups are in the works, its just going to take a bit to write out the lists

jabbink:
Not sure if a Python user interface is that userfriendly. Users of this bot already have issues getting a correct Java version installed.

Therefore I proposed (in the issue) that maybe an online hosted form would be more user friendly.

ZingBallyhoo:
@jabbink I hear what you are saying, and yes, a web settings page would be better
The way to get around the Python requirement would be to pre-compile the files with Pyinstaller. Would need to be done for Linux, Windows and Mac separately.
Also the Tkinter GUI doesn't look too horrible, but it does definitely not look pretty at all, considering it is just a grey box
[Here](http://i.imgur.com/jg4B3Sd.png) is an updated screenshot

ZingBallyhoo:
The Whole GUI looks a lot neater now, no more rough edges.
Theme changing support coming also going to move "save" and "reset" buttons into a frame

ZingBallyhoo:
@jabbink Should I continue developing this, or just leave it in favour of the planned settings webpage?

Anyway, here are things I added:
The UI is a lot more user friendly now, It has some (Not a lot due to limitations in Tkinter) colour now.
You can now change the theme, there are 7 different themes that are built into the GUI toolkit.
Also fixed a major bug where the file didn't save properly.

[Here](http://i.imgur.com/yjqwF7u.png) is the latest screenshot

ZingBallyhoo:
Cross platform building is in the works, even if this is never added to this project, it will be used for personal projects in the future.
At the moment I cant get travis-ci to detect the os without crashing :(

langerhans:
Hm, imho this might be better suited for its own repo? That would also give you the freedom to commit what and whenever you want. Thing is, if it's included here then people will expect it to work all the time. Given that the config changes with each release right now this might not always be the case cause atm it's probably just you who would work on this, right?
But that's just my 2 cents, ultimately that's @jabbink's decision. Still nice work tho 👍

ZingBallyhoo:
@langerhans Thats a good idea. In fact I already did it to try and setup the travis-ci building. I'm currently trying to setup Mac OS building, and when that is done, @jabbink's worry about people having to install another language on their computers. I'm also considering adding multi language support, should I bother doing it?

The separate repository can be found here:
https://github.com/ZingBallyhoo/PokemonGoBot-ConfigTool

ZingBallyhoo:
OK so multi-platform building should be working now. I wont be able to test the Mac OS build until I get my shiny new Mac OS virtual machine up and running. But I can confirm executable files are being built for Linux, Windows and Mac OS

ZingBallyhoo:
@jabbink I need your input.
@langerhans Suggested that this config tool is in a separate repository where compiled executable files are stored (No Python required to run, just download the file for your OS). And then people could simply be linked to that repo (from this repo's release notes?), go to the releases tab and get the files they need. This would make issue tracking much easier. Also a new pull request would not be needed for every update.

The only unfortunate things is that Mac OS has to be in a seperate repository (Because of Travis-CI deployment). But the user could just be linked to the Mac Repo from readme.md in the Windows and Linux repository.

Linux, Windows and Mac versions tested and definitely running without Python installed.

Also this is quite a bit more user friendly than it was in the first version, it looks so much better now.

Here are the two repositories:
[Windows & Linux](https://github.com/ZingBallyhoo/PokemonGoBot-ConfigTool)
[Mac OS X](https://github.com/ZingBallyhoo/PokemonGoBot-ConfigTool-Mac)

luismiguelsec:
how can i compile the source code?

ZingBallyhoo:
@luismiguelsec To compile follow what it says to do in readme.md
You might as well just run a version from the releases tab, becuase I havn't done any work in a while

ZingBallyhoo:
@luismiguelsec The info in under: "Running from source"

PR #487, opened 2016-07-27
re-store pokestops near player position

jabbink:
**Changes made:**

* re-store pokestops near player position (gui allows walking far away from the start point and the bot won't see the pokestops otherwise)
* also increase walking speed again
* also handle Ptc login errors better


sodanakin:
Did you just increase the default speed or the speed somewhere in the code? Wondering about this because I used to run this bot at values between 50.00 to 100.00 speed without any softbans. And if there's some multiplier increase now I'd might want to turn it down a bit ;p

jabbink:
Changed the default speed to a lower value yesterday, because it looked like the server was not recognizing the distance walked properly anymore. Appeared to be a temporary issue and I set it back now to a tested speed that is optimal for egg hatching

image

jabbink:
Introduced (default: off) property to disable this behavior.

PR #496, opened 2016-07-27
Changed profile logging color from White to Magenta

ReignOfComputer:
Not sure if Magenta is being used for something else - if it is I didn't
see. White isn't very visible especially when next to the normal text
outputs. Profile updates should have priority colors. Just a suggestion.


sirsean:
LGTM 👍

Issue #328, opened 2016-07-25
Walk back to the starting point


pashorizer:
Any plans on having a better walking algorithm? Since the bot walks from pokestop to pokestop which ultimately makes the bot leave the starting area e.g. central park is full of pokemons but not that many pokestops.

I thought about smth like after 5 minutes head back to the first pokestop you encountered regardless of distance.

TaylorZaneKirk:
I think that would be pretty neat. I would definitely find it useful, at least. And it would help with an issue I sometimes run into. It seems that if the bot 'walks' to a pokestop that, once there, the bot can no longer reach any other pokestop and the bot get stuck and must be restarted.

Rofii:
An option to define a radius from the starting point which the bot will not exceed would be better, in my opinion anyway. Simply tell it not to go farther than X meters in any direction from your starting coordinates.

pashorizer:
@TaylorZaneKirk Yes the bot is coded like that. It searches the nearest Pokestop and goes there. Every Pokestop has a timeout which indicates if the bot should go there again or not. So what happens is that there is a endless circle of pokestops which are out of the starting area e.g. central park

![repeat](https://cloud.githubusercontent.com/assets/5606097/17114963/7e287e58-52b1-11e6-8a99-bdc4c8ab8d05.png)

My proposal is that there is a new config `backToStart=5min`s and the red line will be followed back to the starting point while catching everything on the way.

sargunster:
I would love to have both a back to start option and a radius limit option.

jabbink:
> if the bot can no longer reach any other pokestops, the bot gets stuck and must be restarted.

This bot is intended to be used in areas that actually have pokestops (cities). Currently it runs circles through all pokestops in a 3km * 3km radius, if you run out of pokestops you can get within those 5 minutes, why are you using a bot that is made to gain XP quickly in this area in the first place? Pick a city and run the bot there.

TaylorZaneKirk:
Totally understandable. Central park is just a really good spot with not THAT many stops. I do take your point, however.

pashorizer:
This somewhat solves it partially https://github.com/jabbink/PokemonGoBot/pull/334

Sieberkev:
#334 improves it a lot (even seen a lot more "new" pokestops), but after a while it got stuck in a loop again...

Now I just restart it manually if I notice it, but a backToStart of 30m or 1h would be lovely :angel:

jabbink:
[...]

PR #518, opened 2016-07-27
Add max distance from start option

jabbink:
**Fixed issue:** #328 

**Changes made:**

* Add max distance from start option


sphinxid:
the example of the distance is "1". one as in 1 meter?

PR #521, opened 2016-07-27
Make the socket listen on 0.0.0.0

langerhans:
**Fixed issue:** Can't use WebGUI when not running from localhost.

**Changes made:**

* Bound to 0.0.0.0 instead of 127.0.0.1

This removes the socket listening limitation. Otherwise you will only be able to run this interface when running the bot on localhost.


jabbink:
As this is an early test version of the Web GUI I'll allow it.

@ users: make sure your firewall is set up correctly.

ColeB27:
@jabbink do you think it could be permanent?

pilotchute:
@ColeB27 It should not be. It is not a secure way of allowing the action.

PR #522, opened 2016-07-27
move isTooMany logic

jabbink:
**Changes made:**

- `obligatoryTransfer` pokemon should always be released
- non-obligatory & non-ignored pokemon should go through the whole logic
-  `ignoredPokemon` should never be released (like `isFavourite` ones)"

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/ReleasePokemon.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/ReleasePokemon.kt
@@ -38,7 +38,7 @@ class ReleasePokemon : Task {
                     val ivPercentage = pokemon.getIvPercentage()
                     // never transfer highest rated Pokemon (except for obligatory transfer)
                     if (settings.obligatoryTransfer.contains(pokemon.pokemonId.name) || index >= settings.keepPokemonAmount) {
-                        var (shouldRelease, reason) = pokemon.shouldTransfer(settings, pokemonCounts)
+                        val (shouldRelease, reason) = pokemon.shouldTransfer(settings, pokemonCounts)
 
                         if (shouldRelease) {
                             Log.yellow("Going to transfer ${pokemon.pokemonId.name} with " +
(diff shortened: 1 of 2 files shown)


src/main/kotlin/ink/abb/pogo/scraper/tasks/ReleasePokemon.kt
```diff
@@ -38,14 +38,7 @@ class ReleasePokemon : Task {
...
                     if (settings.obligatoryTransfer.contains(pokemon.pokemonId.name) || index >= settings.keepPokemonAmount) {
-                        var (shouldRelease, reason) = pokemon.shouldTransfer(settings)
-
-                        if (!shouldRelease) {
-                            if (isTooMany(settings, pokemonCounts, pokemon)) {
-                                shouldRelease = true
-                                reason = "Too many"
-                            }
-                        }
+                        var (shouldRelease, reason) = pokemon.shouldTransfer(settings, pokemonCounts)
```
sirsean: These could be `val`s now that we're not changing them, right?

jozsi:
@jabbink:

I prototyped some changes (commented):
```kotlin
// Move this outside the if so it will always update `pokemonCounts`
// and because the `obligatoryTransfer` check is redundant
val (shouldRelease, reason) = pokemon.shouldTransfer(settings, pokemonCounts)
// We could even move the `keepPokemonAmount` condition into `shouldTransfer`
// as it doesn't need to know the `index`, cause it has `pokemonCounts` for that
if (shouldRelease && index >= settings.keepPokemonAmount) {
    Log.yellow("Going to transfer ${pokemon.pokemonId.name} with " +
            "CP ${pokemon.cp} and IV $ivPercentage%; reason: $reason")
    val result = pokemon.transferPokemon()
    if (result == Result.SUCCESS) {
        ctx.pokemonStats.second.andIncrement
        ctx.server.releasePokemon(pokemon.id)
        ctx.server.sendProfile()
    } else {
        Log.red("Failed to transfer ${pokemon.pokemonId.name}: ${result.name}")
    }
}
```

PR #526, opened 2016-07-27
Stop socket server when app stops.

apottere:
Fixes "Address already in use" when stopping/starting quickly.


shilch:
Thank you, During development I always had to kill the bot because of this AddressInUse.

PR #532, opened 2016-07-27
Implemented walking following streets with option to disable

pr0ves:
Finally made it.
Needs some test regarding the timeout value, default to 200L.
Could be too small, making the bot go faster than it should.
routingmadafaka


Simpuhl:
Are we able to click anything on screen to force bot to walk there or is this just all to view things easier than reading the log?

pr0ves:
The bot already implements a way to force go somewhere, with my commit it will possible to go wherever you want, but only after it will go to an already decided pokestop, like: the bot is going to a randomly chosen pokestop, you right click to the location where you want it to go, the bot will go first to the pokestop and then to the chosen location.

Simpuhl:
@pr0ves 

That's awesome.

Question though, does the pokemon that appear on the screen, are those the ones that are in the area using a scan or are they the ones that you've already captured? It would be cool to see the scanned area, so that if you see something rare then you can click it.

build.gradle
```diff
@@ -42,9 +42,12 @@ dependencies {
 
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     compile "com.corundumstudio.socketio:netty-socketio:1.7.7"
+
+    compile group: 'org.json', name: 'json', version: '20160212'
```
jabbink: maybe use Moshi, which is already used as dep by the API (decreases file size and download time)

config.properties.template
```diff
@@ -25,6 +25,9 @@ random_next_pokestop_selection=5
 # Meters per second
 speed=2.8
 
+# Should the bot follow the streets (true) or just go directly to pokestops/waypoints
+follow_streets = true
```
jabbink: and maybe make the default false, some people use this bot for max xp/hour; I guess this doesn't result in more xp/hour

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
```diff
@@ -36,6 +36,7 @@ class Settings(val properties: Properties) {
     val token = { properties.getProperty("token", "") }
 
     val speed = getPropertyIfSet("Speed", "speed", 2.778, String::toDouble)
+    val shouldFollowStreets = getPropertyIfSet("Should the bot follow streets", "follow_streets", false, String::toBoolean)
```
jabbink: lol, here it is false already

pr0ves:
Those are the caught one, another account would be needed to scan the area

Simpuhl:
@pr0ves 

Will we have an option to add two accounts?

pr0ves:
Don't know, we'll see.

build.gradle
pr0ves: I just need two functions from org.json; with Moshi you need a class which must have the same properties as the json, then use Moshi.Builder().build() etc...
It would complicate things. Besides the org.json jar it's only 50KB,

accident12123:
What happens when there is no street path to the destination?

pr0ves:
It depends, in cities the pokestops are always near streets so even if it's not reachable, the bot will go to the nearest point on the street. 
Anyway I will add a check that will use the direct walking function if there's no path to a point.

accident12123:
sounds good, areas with rivers and lakes its not uncommon for a pokestop that is on an island and not walkable.

pr0ves:
#599

Issue #505, opened 2016-07-27
[Suggestion] Export pokemon list to CSV/SQL


rannmann:
It would be cool to have a command-line option or config option to keep a database or database-like file up to date with the stats of all your pokemon.  Below are the things I'd love to have access to:

- Pokemon Name
- Pokemon Id (type)
- Favorited Boolean
- CP
- IV
- Stamina (HP)
- Move 1
- Move 2
- iAttack
- iDefense
- iStamina
- cpMultiplier

Of less importance but would still be neat

- Favorited Boolean
- Location acquired
- Height
- Weight

This would allow external applications to manipulate the data,.  The built-in interface for viewing pokemon requires the loading of each individual 'mon, and it's hard to pick the best option out for the one you intend on powering up and leveling.  Additionally, it would allow for a future import option to rename pokemon with the matching stats to prevent accidental deletion.

pilotchute:
@rannmann Yes. All. The. Yes.

Sieberkev:
I have added a PR #536 that will export a bunch of data to a CSV file on each ProfileLoop. If you could check it out?

I think the renaming/favoriting/evolving/... features can be added to the awesome Web GUI system later, but this can be used as a base to parse to a DB anyway :smile:

langerhans:
Added in v0.5.0-alpha3, thanks @Sieberkev !

PR #536, opened 2016-07-27
Export CSV (with playerstats/pokebank overview) every ProfileLoop

Sieberkev:
**Fixes issue:** #505

**Changes made:**

* Added configuration to optionally export a CSV every ProfileLoop (default = false)
* Added task ExportCSV, export a bunch of stats to a CSV file

I don't have a lot of experience in this language, but I found this issue an easy one to start with so feel free to give feedback :smile:


apottere:
Kotlin has a multiline string that you can use, e.g.
```kotlin
writer.write("""1,2,3
4,5,6
7,8,9""")
```

Emuuung:
Tested and working very well :+1:

Sieberkev:
@apottere I read that the indent handling of the multiline writer is kinda [annoying](https://discuss.kotlinlang.org/t/multiline-string-literal-indent-handling-is-unconvient/497), also the println makes sure the line endings are always correct (I think)?

I have to wait on a [fix](https://github.com/Grover-c13/PokeGOAPI-Java/issues/302) in the Java API for getting the correct weight.

Sieberkev:
If there are no more issues or wanted functionality, I guess this is ready for now? :smile:

Sieberkev:
The friendly people from the PokeGoAPI-java repo [implemented](https://github.com/Grover-c13/PokeGOAPI-Java/pull/309) the getWeightKg() function in the PokeGOAPI :tada:

... but for the life of me I can't get it working locally at all.

If I pull the most recent PokeGOAPI (`.../pogobot/lib/PokeGOAPI-Java ((2fac026...))` so far so good) and try to get it working (`git submodule update --init --recursive`, `./gradlew build`), I get compile errors (`cannot find symbol .getCandy()` if I try to build the PokeGOAPI-Java directory, `Could not find ...svarzee:gpsoauth-java:v0.2.0` if I try to build the PokemonGoBot directory) that I can't fix with my limited knowledge of them meddling Java dependencies...

Sieberkev:
After merging upstream, everything works again :tada: 

If no one has any more remarks, I guess it can be merged?

Sieberkev:
There, 43 columns of data from your Pokemons spit out every ProfileLoop :smile:

pr0ves:
This is how i get the csv
export
Shouldn't they be the values in other cells?

Sieberkev:
Yeah, it's because of USA/UK CSV (which uses "," as separator) vs European CSV/DSV (which uses ";" as separator). On my excel it doesn't work with "," either, I have to do a text import for it to work correctly...

Standards right, why even bother ¯\_(ツ)_/¯

I can put a separator char as configuration, that's how a lot of other scripts fix that issue...

pr0ves:
Isn't there another way to fix this?
Wouldn't like another option to the settings.

Sieberkev:
It depends on the (locale) settings used in the OS/program of the user that does the import. To avoid a cluttered configuration I made it as described in the [RFC](https://www.ietf.org/rfc/rfc4180.txt) and use Data -> "Import from text" in MS Excel (which is a little workaround, but it works).

Other existing JAVA libraries that do CSV stuff use the same default settings too, and use additional configuration to override them.

Sieberkev:
jabbink:
> I think semicolon separated is easier for most

I could add it as optional configuration if you want it, maybe instead of a boolean a string "no"/"comma"/"semicolon", so it stays one configuration key. Or make it even more exporty with options: ​*empty = off*​,CSV (comma),DSV (semicolon),JSON,XML,...

I guess I could put that as extra functionality in a new PR?

Sieberkev:
@pr0ves 
> Isn't there another way to fix this?

I changed the "export_csv" boolean key to one "export" key where you can put the format of the export (currently supports none, CSV, DSV).

I also put all the data in an ArrayList<Array<String>> thru a CSVWriter class so if someone want other formats/exports (JSON/XML/XLS/XLSX/SQLite/...) one would only have to write an array -> export for the desired configuration :)

And that all with only one export config key :tada:

Samoth1980:
Sorry for the question, but i've set it like "CSV" but where the file is saved?

Sieberkev:
In the same directory where you start the script :smile:

Samoth1980:
Thank but i don't find it, i need to modify the configuration with only "csv" or some other things?

PR #544, opened 2016-07-28
Separate settings from config.properties data and fix google login without username/password.

apottere:
Right now the properties are parsed from config.properties in the construction of the Settings object - this means Settings can only ever come from properties files.

This change decouples the Settings and the config.properties file, adding a factory method that creates a Settings instance from a config.properties file like before.

This allows us to store settings files anywhere - not just properties files on disk.

- [x] Separate settings
- [x] Fix google passwordless login


apottere:
Note that this doesn't change the way the bot works right now - config.properties is parsed exactly as it was before.

apottere:
Per the recent changes to login, this PR now also re-adds an option for doing token authentication with Google.

pr0ves:
I'm ready to merge this, do you need to add something else?

PR #549, opened 2016-07-28
Add responsive template and option in config for WebGui

jimzrt:
I like the standard WebGui template but it is not usable on my mobile phone, so I modified the other one to make it responsive with a few changes to refine the look.
Drag/Drop-Support is lost, so I made it optional via config (default is false).

pokemon go bot gui desk
pokemon go bot gui


apottere:
Can it just toggle to responsive based on the screen size?  Seems silly to have a backend setting for this.

jimzrt:
well, the toggle is just to use a different template.
Would be more work to modify the other template to be responsive, but I didn't want to overwrite because it looks different  and choice is always good.
Could by extensible to offer multiple templates in the future, maybe.

langerhans:
I like that one, nice work! I'm currently looking into introducing a real web framework instead of just that webserver so we could add some user agent type switching and transform this config option to always use the responsive template.

kasagon:
I like this GUI, but I have one thing to add to the indexResponsive.html file, Is it an idea to change the @media tag, starting at rule 247 to:

        @media (max-width: 991px) {

            .panel {
                min-height: 50vw;
            }
			
	        [class*="col-"] {
				padding: 2px;
			}
        }

This will make the padding of the windows smaller if you watch the webpage on a small screen.

kasagon:
Would it also be possible to show the exact IV´s of the Pokemon in the Pokebank? (Stamina, Attack, Defense)?

I guess some of the kotlin need to be changed for that though, but it can be handy.

langerhans:
I just opened #589 to re-work the architecture of the gui a bit. If we follow through with this all the way, we would be a lot more flexible also with serving a mobile freindly version dynamically. Also would de-duplicate the code a lot.

jabbink:
Great work, please see #728 

I'm very tempted to move the UI out of this repo and host it externally (eg GitHub Pages), that way multiple UI's can peacefully co-exist and users can pick whatever they want. For that to work we need to better define the current web socket behaviour (eg. versioned url's and UI's may request updates on specific items, so some mobile UI/app that only shows the status of your x bots wouldn't receive all map updates, and versioned, so new UI's can use newer functions while old UI's still receive the expected data)

swtrpeter:
How can I add this ? Love the way it looks

jabbink:
@jimzrt if you like to add this, make a pr to the `gh-pages` branch please (either directly push into it, or add it as a submodule of your own repo (like the RocketTheme))

Issue #563, opened 2016-07-28
Bot doesn't show when you don't have any Pokeball, and doesn't resume catching after you got some Pokeballs.


kasagon:
#### Log:

28 jul 04:30:53: Caught a SPEAROW with CP 100 and IV 57%
 Stamina: 2 | Attack: 12 | Defense: 12 | IV: 26 (57%): [200x XP, 3x Candy, 100x Stardust
28 jul  jul 04:30:53: Going to transfer SPEAROW with CP 100 and IV 57%; reason: CP < 500 and IV < 80% 
28 jul 04:30:53: Resuming walk.
28 jul 04:30:57: Getting map of (XX.XXXX, XX.XXXX)
28 jul 04:31:02: Getting map of (XX.XXXX, XX.XXXX)
28 jul 04:31:07: Getting map of (XX.XXXX, XX.XXXX)
28 jul 04:31:12: Getting map of (XX.XXXX, XX.XXXX)
28 jul 04:31:12: Pausing to catch pokemon...
28 jul 04:31:13: Resuming walk.
28 jul 04:31:14: Pausing to catch pokemon...
28 jul 04:31:15: Resuming walk.
28 jul 04:31:15: Pausing to catch pokemon...
28 jul 04:31:17: Resuming walk.
28 jul 04:31:17: Pausing to catch pokemon...
28 jul 04:31:17: Getting map of (XX.XXXX, XX.XXXX)
28 jul 04:31:20: Resuming walk.
28 jul 04:31:20: Pausing to catch pokemon...
28 jul 04:31:21: Resuming walk.
28 jul 04:31:22: Pausing to catch pokemon...
28 jul 04:31:22: Resuming walk.
28 jul 04:31:22: Pausing to catch pokemon...
28 jul 04:31:22: Getting map of (XX.XXXX, XX.XXXX)
28 jul 04:31:23: Resuming walk.
28 jul 04:31:23: Pausing to catch pokemon...
28 jul 04:31:25: Resuming walk.
28 jul 04:31:25: Pausing to catch pokemon...

The bot used it last Pokeball to catch the Spearow, after that the bot was pausing, and resuming the whole time, what caused to move the character really slowly, without any message that you don't have any Pokeballs.

After the bot got some Pokeball's again from a Pokestop, it didn't resumed the catching, but ignored the fact that it got Pokeballs again.

## **Steps to reproduce:**

1. Have no Pokeballs.
2. Let the bot walk.

## **Expected behavior:**
[...]

PR #574, opened 2016-07-28
Stop only if Pokeballs are available & fix for not walking anymore

pashorizer:
Should fix https://github.com/jabbink/PokemonGoBot/issues/563

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/Walk.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/Walk.kt
@@ -92,19 +92,18 @@ class Walk(val sortedPokestops: List<Pokestop>, val lootTimeouts: Map<String, Lo
             // incase api does not respond ; just continue
             var walking = true
             // don't run away when there are still Pokemon around
-            if (walking) {
-                if (ctx.api.inventories.itemBag.hasPokeballs() && bot.api.map.getCatchablePokemon(ctx.blacklistedEncounters).size > 0 && settings.shouldCatchPokemons) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/ink/abb/pogo/scraper/tasks/Walk.kt
```diff
@@ -77,22 +78,24 @@ class Walk(val sortedPokestops: List<Pokestop>, val lootTimeouts: Map<String, Lo
...
-                    walking = false
-                    Log.normal("Pausing to catch pokemon...")
-                } // Else continue walking.
-            } else {
-                if(bot.api.map.getCatchablePokemon(ctx.blacklistedEncounters).size <= 0) {
-                    walking = true
-                    Log.normal("Resuming walk.")
-                } // Else continue waiting.
-            }
+            if (ctx.api.inventories.itemBag.hasPokeballs()) {
```
apottere: This line should be further below - in the `if` statement that checks if walking should pause.  Otherwise it could cause the bot not to start walking again when you use your last pokemon.

pashorizer: True. Gonna fix it later

pashorizer:
@apottere Should be fine now

pashorizer:
@apottere @pr0ves all set

pashorizer:
@Boraida Yup, you are right. It got pretty late yesterday. Initially I did it with a `try-catch` block to avoid api exceptions, but realized it's simpler that way.

PR #581, opened 2016-07-28
Added kiddy bumper to front page.

Agronis:
Changes are to that of the Front page.

Adding a buffer of READ THIS FIRST

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,19 @@
 
 [![Build Status](https://travis-ci.org/jabbink/PokemonGoBot.svg?branch=develop)](https://travis-ci.org/jabbink/PokemonGoBot)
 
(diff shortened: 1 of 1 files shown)


rirze:
Nice! Could you also add the java1.8 requirement to the directions for compiling from source (this took me a while to figure out when I was building)

Agronis:
I feel like if you're compiling from source - you're aware of what you're doing.

Otherwise, why not just use the jar release?

rirze:
It's nice to list dependencies without expecting people to know what to have before building. Regardless, it's just a tiny point, feel free to disregard.

Agronis:
Not disregarding.. I'll be sure to break it down soon.

Issue #534, opened 2016-07-27
[Web GUI] Add name of trainer in title


jabbink:
**Description:**
Makes it easier when you have many tabs with many bots to find a specific tab

Emuuung:
@jabbink Do you mean adding the name of your PTC account or your google email to title?

kasagon:
@Emuuung I think he means your trainer name, This should be not that hard to implement with JavaScript (document.title code), plus there is already a variable of the trainer name.

Emuuung:
Ah, I was confused, you mean the title of the web browser when selecting the web GUI, I thought you meant the Profile window, because it already has the name of the trainer.

langerhans:
This works in the RocketTheme UI.

PR #589, opened 2016-07-28
[WIP] Re-work web ui architecture

langerhans:
**Fixed issue:** Mostly not a fix, but still added a fix for #534 

**Changes made:**

* Introduced Spark as lightweight web framework to serve the GUI and maybe more later (REST API?)
* Added trainer name to page title

**To dos (Yet this can be merged on its own)**

- [ ] Split up the huge HTML into the template, JS and CSS file for reusability.
    - [ ] Introduce a real templating engine
    - [ ] Remove the String replacements with actual models
    - [ ] Move all assets to respective folders
- [ ] Migrate to Spark's Jetty WebSocket

**Discussion about To dos**
About the templating: There are tons of options for a templating engine. Spark supports several out of the box, see a list here: http://sparkjava.com/documentation.html#views-templates
I personally have used Pebble before, which worked fine. Thymeleaf comes with the benefit that previews still render as normal HTML but the learning curve is steep. Velocity is recommended often and very mature.
To be honest, we probably won't need that many features for now, but we should keep the options open if the web gui gets bigger.

About the websocket: Spark comes with Jetty which comes with a websocket server. Problem is we use Socket.IO right now which is kinda convenient but also a bit special, so not easy to adapt. The only real advantage of moving would be one less dependency and being able to use the same port as the website is running on also for the websocket. But not sure if that's worth it.

### DIFF
--- a/src/main/resources/ink/abb/pogo/scraper/gui/index.html
+++ b/src/main/resources/ink/abb/pogo/scraper/gui/index.html
@@ -10,7 +10,7 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <title>Pokemon Go Bot GUI</title>
+    <title>{{username}} - Pokemon Go Bot GUI</title>
 
     <link rel="stylesheet" href="https://ajax.googleapis.com/ajax/libs/jqueryui/1.11.4/themes/smoothness/jquery-ui.css">
     <link href="https://fonts.googleapis.com/css?family=Lato" rel="stylesheet">
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ dependencies {
 
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     compile "com.corundumstudio.socketio:netty-socketio:1.7.7"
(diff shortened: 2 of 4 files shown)


apottere:
Do we need backend templates?  This seems like a very good fit for a SPA.

src/main/resources/ink/abb/pogo/scraper/gui/index.html
```diff
@@ -10,7 +10,7 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <title>Pokemon Go Bot GUI</title>
+    <title>{{username}} - Pokemon Go Bot GUI</title>
```
shilch: You can change the title with JavaScript: `document.title = "Page title" `

langerhans: Yeah I know, but the nick will not change dynamically I think?

shilch: Yes, it will not change but data should be added in the client and not Server (my opinion)

langerhans:
Closing in favor of #622

PR #622, opened 2016-07-29
Spring boot (take 2) with multiple bot capabilities.

apottere:
Re-add spring-boot in an unobtrusive (read: easy to merge) way, so there aren't as many merge conflicts as before.  Also re-adds functionality for defining multiple bots in JSON files.


apottere:
I can't actually test this fully right now because I can't get past login (on develop or this branch), but it seems to work up to login.

apottere:
This also uses a `CommandLineRunner` to avoid getting a huge stacktrace when the bot fails.

sodanakin:
Hold up lol, can we add this after we add a fix for the recent server patch? This stuff legit broke the whole thing for me last time x)

apottere:
It only broke the whole thing because the PR languished and had to be merged 6 times.  I tried to do my best, but I had changed too much at once.  This changes much less, and shouldn't have an effect on how the bot works normally.

sodanakin:
Ah alright whelp, lets hope he gets on soon anyways. Bots choking x)

Emuuung:
@apottere This isn't related to the latest throttle server patch?

src/main/resources/ink/abb/pogo/scraper/gui/index.html
```diff
@@ -538,7 +538,7 @@
             }
         }
 
-        var socket = io.connect('ws://' + window.location.hostname + ':{{socketPort}}');
+        var socket = io.connect('ws://' + window.location.hostname + ':8080/socket');
```
langerhans: Did you want that to be hardcoded?

build.gradle
```diff
@@ -44,8 +50,14 @@ dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     compile "com.corundumstudio.socketio:netty-socketio:1.7.7"
 
+    compile 'org.springframework.boot:spring-boot-starter'
+
+    compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: jacksonVersion
+    compile group: 'com.fasterxml.jackson.core', name: 'jackson-core', version: jacksonVersion
+    compile group: 'com.fasterxml.jackson.core', name: 'jackson-annotations', version: jacksonVersion
+    compile group: 'com.fasterxml.jackson.module', name: 'jackson-module-kotlin', version: jacksonVersion
```
langerhans: We already have a transitive dependency on Moshi. And another PR will (did?) add org.json. If we include Jackson now we'll have three JSON libs.

langerhans:
This clashes a bit with #589 but I'd be totally fine with basing my work there on Spring instead. Have another bot running on it as well. I guess we should discuss the overall way forward architecture wise when @jabbink is back.

Agronis:
+1 For SBoot

apottere:
@Emuuung no, I think the best place for that to be fixed is in the API itself, but *shrug*

src/main/resources/ink/abb/pogo/scraper/gui/index.html
apottere: Nope, this was accidental.

apottere:
@langerhans Yeah, I saw that PR right after I opened this one.  I'd much rather use boot since it gets us more than just a webserver (namely DI).  We can still add GSP/Thymeleaf if that's the way we want to go, and REST APIs are dead simple with MVC.  Even websockets are a breeze with STOMP, if we're reworking the UI.

apottere:
@Agronis I don't think file size is an issue, as most people are just running the bot on their own machine.  Build time is more important, and I don't think this adds much.

build.gradle
apottere: Yeah, I saw those dependencies.  I can definitely try to use one of those, or change the other usages to jackson, but I didn't know if either of those libraries could do the same @JsonTypeInfo stuff in the Settings.

langerhans: Well, seems like you're only using that to serialize the credentials, which I'd argue shouldn't be returned by the API anyway ;D

langerhans:
Been running this for a bit now and works fine on my end. I'd suggest splitting the addition of the UI into its own PR when it's finished. No use in merging this until then, end then it's quite large again. I'm more a fan of compact, testable PRs.

build.gradle
apottere: Right - I'm serializing them to disk, not to http.

apottere:
@jabbink @langerhans I removed the UI boilerplate - we can do that somewhere else.  If you want to test this, it should be good to go.

apottere:
With the latest commit, you can do the following if you're using the JSON config:

```
curl localhost:8080/api/bot/<bot-name>/stop
curl localhost:8080/api/bot/<bot-name>/start
curl localhost:8080/api/bot/<bot-name>/unload
curl localhost:8080/api/bot/<bot-name>/load
curl localhost:8080/api/bot/<bot-name>/reload
```

config.properties.template
```diff
@@ -23,7 +23,7 @@ random_next_pokestop_selection=5
 
 ## Walking
 # Meters per second
-speed=2.8
+speed=2.778
```
jabbink: why reset these default values again 1/3

config.properties.template
```diff
@@ -39,7 +39,7 @@ auto_fill_incubator=true
 
 ## Items drop
 # Should the bot EVER drop items (true/false)
-drop_items=true
+drop_items=false
```
jabbink: why reset these default values again 2/3

config.properties.template
```diff
@@ -78,7 +78,7 @@ display_pokestop_name=false
 
 ## Pokemon Transfer
 # Should the bot auto transfer duplicate pokemon
-autotransfer=true
+autotransfer=false
```
jabbink: why reset these default values again 3/3

src/main/resources/banner.txt
```diff
@@ -0,0 +1,13 @@
...
+\      __    \    '-.  | /   `.  ___    |    \/    |   '-.   \ |  |   \  \ .---. .-' /
+ \.    \ \   |  __  |  |/    ,','_  `.  |          | __  |    \|  |    '. '     `\_.'
+   \    \/   /,' _`.|      ,' / / / /   |          ,' _`.|     |  |      |(),()  |     ,
+    \     ,-'/  /   \    ,'   | \/ / ,`.|         /  /   \  |     |      (  __   /   .' \
+     \    \ |   \_/  |   `-.  \    `'  /|  |    ||   \_/  | |\    |     .''.___.'--,/\_,|
+      \    \ \      /       `-.`.___,-' |  |\  /| \      /  | |   |    {  /     \   }   |
+       \    \ `.__,'|  |`-._    `|      |__| \/ |  `.__,'|  | |   |     '.\     /_.'    /
+        \_.-'       |__|    `-._ |              '-.|     '-.| |   |      |'-.-',  `; _.'
+                                `'                            '-._|      |  |  |   |`
+                                                                         `""`""`"""`
```
jabbink: :+1: :heart:

jabbink:
Will look into this tomorrow before merging that UI branch of mine.

apottere:
The values are reset because that's what the test you added needs.

jabbink:
The test needs consistent values between Settings.kt and the template. The speed of 2.8 is tested to be the most optimal for hatching eggs. The other values... I don't know. At some point they were suddenly so most behavior was disabled by default (causing many issues and imho default settings shouldn't suddenly change between versions, otherwise old users are confused as well)

apottere:
I can change the settings defaults in Settings.kt if that's what you're getting at.  Were those default before this merge?

jabbink:
Yes, well they were up until the weekend, will change it when merging.

jabbink:
Merged in b7fcc97b7a10533da5d3c745af4b32ffc127c2e3

jabbink:
@apottere http://localhost:8080/api/bots reports the credentials. Maybe that should be stripped out (open networks; bad firewall; rogue UI's that pull credentials that way)

apottere:
Yeah, we should probably remove that on the way out.  Good catch.

smileytechguy:
Can the API calls be made through a socket?  I'm running into cross origin control problems

PR #624, opened 2016-07-29
Limit rethrow ball speed

zenwaichi:
to prevent server request spam bursts


langerhans:
Thanks for the contribution, but this has been fixed by the upstream api update that deals with request limting.

PR #625, opened 2016-07-29
Delay task execution

zenwaichi:
to prevent bursts of consecutive api calls. 
With the pokeball rethrow commit it removes 95% of the loop errors ( Disabling profile I went error free for over 30 minutes)

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
@@ -174,6 +174,7 @@ class Bot(val api: PokemonGo, val settings: Settings) {
     }
 
     fun task(task: Task) {
+        Thread.sleep(300)
         task.run(this, ctx, settings)
     }
 }


destilla:
When I run this with your pokeball rethrow I get endless walking loop errors. Did I miss something else that you changed?

Emuuung:
@zenwaichi I see you are adding the same single line to various calls the bot does in various requests like throwing pokeballs and other tasks, I would suggest to make a single PR with all these changes.

zenwaichi:
There's too many places where a single task calls the api several times. I fixed them all on my local repo but I don't think it's code that should be spread due to low quality. Let's wait for either jabbink or the api guys to find a more elegant solution (like wrapping api call internally in a timed queue)

Emuuung:
I've been trying other bots that update profile quite fast, and they had already a task for catching pokemon, dropping items and looting pokestops in a timed matter, so they are operating just fine.

PR #640, opened 2016-07-29
Add Pokévision button for map

smileytechguy:
Added a button for the GUI that will open Pokévision in a new tab with the center of the map's location

### DIFF
--- a/src/main/resources/ink/abb/pogo/scraper/gui/index.html
+++ b/src/main/resources/ink/abb/pogo/scraper/gui/index.html
@@ -97,19 +97,19 @@
             height: 400px;
         }
 
-        #pokemon_list {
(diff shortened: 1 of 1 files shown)


citruz:
Great Idea, but wouldn't it be better to have this as a button above the map instead of an additional window?

smileytechguy:
@citruz with the way I use the GUI this is best way for me.  I can do it on the map if you would rather
screen shot 2016-07-29 at 7 37 51 pm

langerhans:
I mean it's personal preference, but for something so simple, I'd also think it's better on the map. Maybe give the map top bar some padding and add it there.

smileytechguy:
@langerhans Alright, will do

smileytechguy:
Made a new PR with fixes

PR #664, opened 2016-07-30
Smooth walk between Pokestops without stuttering

pashorizer:
Right now, the api takes too long to respond which makes the bot stutter on the way. Originally this solution was to avoid `CATCH_FLEE` failures. So the Bot checked after every step if there is something to catch, but makes it seem like it lags.

Check every 20 steps if something is around.


neox387:
getting about 3 times after a capture:
Encounter failed with result: ENCOUNTER_ALREADY_HAPPENED

langerhans:
So currently the scan radius is 70m. If I didn't do the math wrong, 20 steps in the bot is about 100m. So shouldn't we reduce this to about 14 steps per check? Otherwise it works fine.

pashorizer:
@langerhans I actually choose the value by trial and error and ~ 20 was a sweet spot to not get `CATCH_FLEE` and to stop as few as possible

langerhans:
Okay, in that case i take back my shitty math :D

vamirineni87:
Seeing ENCOUNTER_ALREADY_HAPPENED 5 times for each Pokemon encounter

PR #700, opened 2016-07-31
Added Dockerfile

konstantinj:
**Changes made:**

* No code changes but added files to build and run inside docker container

This makes it at least for me easier to build and run it und various machines.
You'll need at least docker but I've also included a docker-compose.yml which shows how to set properties from the environment.


konstantinj:
Haha, we did exactly the same. Event the implementation with env vars for the properties is the same, even the prefix pogo_ :-D

No, I haven't checked that before because he did it in his own repository.

Only difference is that mine is creating the fatJar and uses this to run.

jabbink:
Why close?

konstantinj:
Because I fucked up the fork ;-)

But I've already recreated. Wanted to wait until you've fixed the API changes before I open it again.

PR #709, opened 2016-08-01
GUI: Added Eggs Window

BerserkerJP:
Added simple eggs window, doesn't update live though (page needs to be refreshed)

screen shot 2016-08-01 at 02 41 48

EDIT: Added percentage


smileytechguy:
could we get a 3x3 table?  and maybe a number/percentage?

jabbink:
Why closed?

PR #728, opened 2016-08-01
Move GUI to GitHub pages

jabbink:
**Changes made:**

* Move UI to GitHub pages: `http://pogo.abb.ink/map.html`
* Remove HTTP webserver from project
* Add console log line to direct to GH Pages


jabbink:
User gets redirected like this

image

apottere:
How is versioning going to work with this?  Will you only be able to use the UI with the latest version of the bot?  Is every version going to be available on the pages domain?

Killberty:
Also, if my Bot runs on my raspberrypi, i won't be able to open the GUI website on my PC, right?

jabbink:
>Also, if my Bot runs on my raspberrypi, i won't be able to open the GUI website on my PC, right?

You can.

>Is every version going to be available on the pages domain?

I thought about that, for now I thought the best way to keep this working is to push version specific UI files and let the bot reference a specific version of the UI.

jabbink:
Ok, for versioning of the UI, the bot now outputs a link with a specific version (the one from `build.gradle`)

Just to not f*ck up with releases, we would now need a test that verifies that that file or symbolic link on the `gh-pages` branch exists.

jabbink:
Reason for this entire movement is getting rid off the UI (that not everybody likes, apparently) and put it somewhere else.

If the web socket communication does not change, any other UI will work as well. So maybe think about any future uses of any UI and implement it in such a way that nothing will break (eg subscribing to events by version, and specific version endpoints will always stay and behave the same way)

JPhix:
@jabbink how would this work when being run on external servers ?

smileytechguy:
i wrote a readme on  how to make a compatible GUI for this in #739

Sieberkev:
Grover just fixed the 1.5 hour login bug in [his latest PR](https://github.com/Grover-c13/PokeGOAPI-Java/pull/442)/commit :)

Magnog:
How do i activate the GUI, is it a separate command?

eruecco87:
@jabbink @smileytechguy:

Just moved the UI y created out of this project, any chance It can get listed in the main repo's README.md ?

https://github.com/eruecco87/PokemonGoBot-RocketTheme

KyleBoyer:
How can I access my map from my phone/separate IP address?

KyleBoyer:
Also if you run multiple bots on one network...

PR #734, opened 2016-08-01
Catch exception when connection fails to fetch WalkRoute from provider

Sieberkev:
**Fixed issue:** When the connection to the WalkRoute provider fails, the bot throws an exception (`java.net.ConnectException: Connection refused: connect`) and it keeps standing still, waiting forever.

**Changes made:**
Catched the exception so an empty string returns.

If I interpreted the code correctly, it should use the normal Walk function instead, is this correct? Please check before merging :D


langerhans:
Nope, if the function returns "", then you'll get an IndexOutOfBoundsException at line 45 in the same file. There should be an additional check for empty strings in that function so it will return an empty list. Then it will just use the regular walk instead.

Sieberkev:
Thanks you @langerhans, I read over the if check too fast :(.

After testing some more possible things that could go wrong with fetching it, I changed the catch to all Exceptions too.

PR #760, opened 2016-08-02
Force evolution before transfer for monsters in obligatory transfer list

el0w:
**Changes made:**

* After capture, if the pokemon is an "obligatory_transfer" (= not wanted for fighting), an evolvution is tried. The evolved pokemon is then re-inspected for transfer. If the evolve fails, the pokemon is transfered.

Simple strategy but it does the job :)


svombergen:
Good feature! I added a new configuration setting evolve_before_transfer myself, to separate from obligatory_transfer. I also didnt use the re-inspection after evolve since these are pokemon I never want to keep anyway.

I wil try to add a feature to configure the amount of pokemon to save up, then bulk evolve them after popping a lucky egg.

alone23:
+1

aerovolvox:
> :compileKotlin
> e: /home/.../PokemonGoBot/src/main/kotlin/ink/abb/pogo/scraper/tasks/ReleasePokemon.kt: (24, 33): Unresolved reference: sortByIV
> :compileKotlin FAILED
> 
> FAILURE: Build failed with an exception.

Why could it be?

PR #776, opened 2016-08-03
WebUI: Internal webserver

thegimliboy:
Hello.
First of all: Thank you for this awesome bot.

Could you please reimplement the internal webserver? I (and probaly many others), would like to connect to a local IP-Adress instead of an external URL. Or could you at least create a new setting which gives the user the ability to select if the bot should host the WebUI, or if just the Websocket should run, to connect via your WebUI? I hope you got what I mean.
Greets Gimli.


jabbink:
Nice pull request, mate.

jabbink:
>would like to connect to a local IP-Adress instead of an external URL.

The UI already needs internet access thanks to Google Maps, so that requirement makes no sense at all.

PR #779, opened 2016-08-03
added multibot explanation

Cypherke:
**Changes made: added multibot configuration**


multibotconfig.json
```diff
@@ -0,0 +1,58 @@
...
+  "spawnRadius" : -1,
+  "banSpinCount" : 0,
+  "transferCpThreshold" : 400,
+  "transferIvThreshold" : 80,
+  "ignoredPokemon" : [ "EEVEE", "MEWTWO", "CHARMANDER" ],
+  "obligatoryTransfer" : [ "DODUO", "RATTATA", "CATERPIE", "PIDGEY" ],
+  "export" : "",
+  "guiPortSocket" : 8001,
+  "initialMapSize" : 3,
+  "version" : "0.5.0-alpha4"
```
jabbink: should be removed I guess, as this is should be generated at runtime. should probably also remove it in the `data class Settings` annotation. @apottere ?

konstantinj:
Hi,
in my opinion it's too much maintenance to keep this working with future versions. It's quite easy to run multiple bots using docker containers. If you like to use the gui you can just open different ports on your host machine.

multibotconfig.json
apottere: Yup, in `@JsonIgnoreProperties`

apottere:
Why was this closed?

Cypherke:
I m guessing it 'll still change some more in the coming days

PR #782, opened 2016-08-03
Update CSVWriter to work with Multibots

Sieberkev:
**Fixed issue:** When running multiple accounts on the same bot, the same "export.csv" file was overwritten by each bot

**Changes made:**

* Write to export_botname.csv file when writing from Bot Main and ProfileLoop (also updated .gitignore with new filename format)

I'm not proud of the ": ProfileLoop" replace to fetch only the botname from the Thread, but I could not find a working AND proper way to do this :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,6 @@ config.properties
 .DS_Store
 
 # exported file
-export.csv
+export_*.csv
 
 bot-settings/
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/Export.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/Export.kt
@@ -108,11 +108,11 @@ class Export : Task {
 
             when (settings.export) {
                 "CSV" -> {
-                    val writer = CSVWriter()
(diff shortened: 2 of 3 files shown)


jabbink:
"proper" way could be to change the CSVWriter class to accept a filename parameter, then in the `Export`-task, which has access to `settings.name`, set the filename.

Sieberkev:
Yes, that too :D But I meant the botname (= Threadname). It starts correctly with "default" when the export is called from [here](https://github.com/jabbink/PokemonGoBot/blob/develop/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt#L94), but the name changes to "default: ProfileLoop" [here](https://github.com/jabbink/PokemonGoBot/blob/develop/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt#L118).

I tried several .replace functions from [Kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/), but none of them had a good return. Or I had to make a very ugly .split(":") / test split.size / use split[0] / etc...

jabbink:
That's what I meant: don't put the thread name in.

Directly take the name from the settings

PR #792, opened 2016-08-03
actually check for pokemon again after encounter

pashorizer:
title


jabbink:
? What does this fix?

pashorizer:
@jabbink After the first encounter during the walk it did not check for a second,third encounter etc.

jabbink:
manually merged

PR #836, opened 2016-08-08
Make pokemon catchable again after transfer when inventory was full

Peyphour:
I noticed that when the pokemon inventory is full, the bot set `settings.catchPokemon` to false.
And when a pokemon is released, the bot do nothing.

**Changes made:**

* Added a field in Context which store the default catchPokemon setting and the inventory status : `pokemonInventoryFullStatus`
* In CatchOneNearbyPokemon, populate `pokemonInventoryFullStatus` with current catchPokemon setting and inventory status before `settings.catchPokemon = false`
* In ReleasePokemon, restore `settings.catchPokemon` to saved value when a pokemon is released and inventory was full

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Bot.kt
@@ -48,7 +48,8 @@ class Bot(val api: PokemonGo, val settings: Settings) {
             AtomicInteger(0),
             Pair(AtomicInteger(0), AtomicInteger(0)),
             mutableSetOf(),
-            SocketServer()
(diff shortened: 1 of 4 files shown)


pashorizer:
Why did you introduce a new variable for switching `settings.catchPokemon`?

Peyphour:
When the pokemon is transferred, you must set `settings.catchPokemon` to the initial value (the value in the config file.). You can't set it to `true` or `false` if it isn't the initial value of the user.
So when the inventory is full, I save the initial value to restore it when a pokemon is transferred.

Peyphour:
Closed, will pull with another branch

PR #840, opened 2016-08-08
Ignore username and password by default in REST API

Peyphour:
Fix for [https://github.com/jabbink/PokemonGoBot/pull/622#issuecomment-237220726](url)

**Changes made:**

* Username and password are not anymore shown in the REST API


jabbink:
What happens when you restart/reload the bot via the API? I think that calls a save on the json file now that I think about it, don't want the credentials to be removed then.

Peyphour:
Isn't it possible to replace them with à dummy string on API calls for bot settings ?

jabbink:
Merging manually

PR #846, opened 2016-08-08
Template file for ptc.json settings

toffaste1337:


Sieberkev:
There's a JSON template in the repository already, but thanks for the effort!

PR #847, opened 2016-08-08
Template file for google.json settings

toffaste1337:


google.json.template
```diff
@@ -0,0 +1,57 @@
+{
+  "latitude" : XX.xx,
+  "longitude" : XX.xx,
+  "credentials" : {
+    "type" : "google",
```
jabbink: `google-auto` is better.

jabbink:
Sorry, I didn't see this PR before. I already added the file in https://github.com/jabbink/PokemonGoBot/blob/develop/json-template.json and referred to it in the FAQ.

Sieberkev:
There's a JSON template in the repository already, but thanks for the effort!

PR #849, opened 2016-08-08
Make pokemon catchable again after transfer when inventory was full

Peyphour:
I noticed that when the pokemon inventory is full, the bot set `settings.catchPokemon` to false.
And when a pokemon is released, the bot do nothing.

**Changes made:**

* Added a field in Context which store the default catchPokemon setting and the inventory status : `pokemonInventoryFullStatus`
* In CatchOneNearbyPokemon, populate `pokemonInventoryFullStatus` with current catchPokemon setting and inventory status before `settings.catchPokemon = false`
* In ReleasePokemon, restore `settings.catchPokemon` to saved value when a pokemon is released and inventory was full

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
@@ -112,8 +112,8 @@ class CatchOneNearbyPokemon : Task {
                     Log.red("Disabling catching of Pokemon")
                     
                     // Save previous value for restore
-                    context.pokemonInventoryFullStatus.first = settings.catchPokemon 
-                    context.pokemonInventoryFullStatus.second = true
+                    ctx.pokemonInventoryFullStatus.first.set(settings.catchPokemon)
+                    ctx.pokemonInventoryFullStatus.second.set(true)
                     
                     settings.catchPokemon = false
                 }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/ink/abb/pogo/scraper/tasks/ReleasePokemon.kt
```diff
@@ -43,6 +43,17 @@ class ReleasePokemon : Task {
...
                             val result = pokemon.transferPokemon()
+                            
+                            if(ctx.pokemonInventoryFullStatus.second.get() && !settings.catchPokemon) {
+                              // Just released a pokemon so the inventory is not full anymore
+                              
+                              // Restore previous value
+                              settings.catchPokemon = ctx.pokemonInventoryFullStatus.first.get()
+                              ctx.pokemonInventoryFullStatus.second.set(false)
+                              
+                              Log.green("Enabling catching of Pokemon")
```
jabbink: Only log this when `settings.catchPokemon == true`

src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
```diff
@@ -110,6 +110,11 @@ class CatchOneNearbyPokemon : Task {
                 Log.red("Encounter failed with result: ${encounterResult.status}")
                 if (encounterResult.status == Status.POKEMON_INVENTORY_FULL) {
                     Log.red("Disabling catching of Pokemon")
+                    
+                    // Save previous value for restore
+                    ctx.pokemonInventoryFullStatus.first.set(settings.catchPokemon)
```
jabbink: Is this necessary as this is already set in https://github.com/jabbink/PokemonGoBot/pull/849/files#diff-4fe846c8e657c44271b65e94cae42312R52

Peyphour:
Yep, i changed that

Issue #427, opened 2016-07-26
[Feature Request] Logging Eggs hatched


oCortex:
It would be cool if all eggs hatched wrote to a log file with all the info, so leaving it overnight we can see what was hatched from eggs. Could be useful for keeping track of higher IV pokes from Hatching.

Agronis:
Let's take this a step further and add a few other details to it.

On bot termination - print the following:

 - Hatched Eggs
 - Distance Traveled
 - XP Gained
 - Evolved (when introduced)
 - Level
 - Current PokeBank

jabbink:
Related to #91

Mixone-FinallyHere:
maybe try:
`$./gradlew run | tee stderr.txt | grep "Hatched" > eggs.txt | grep "Caught" > caughtPokemon.txt`

Issue #250, opened 2016-07-24
Switch to SLF4J for logging


jabbink:
**Description:**
Used to be an overkill, but as it is already in place thanks to Spring, maybe just switch fully over to have a single consistent logging system.

jabbink:
Closed in #894

Issue #91, opened 2016-07-22
Logging to .txt


kasagon:
Would it be possible to log every action to a .txt file? 

I would really appreciate that since CMD doesn't show everything after a while, and since I got errors when I leave the bot running for a while, so I can't watch back what happened.

jabbink:
http://stackoverflow.com/questions/11239924/windows-batch-tee-command might help for now.

For Mac OS X and Linux users: `java -jar pogo.scraper.jar | tee output.txt`

russr123:
i use this as a launch .bat file

--------------------------------------
set LOGFILE=POGO.log
:: Max log file size allowed in bytes before rotation and archive. 1048576 bytes is one megabyte
set LOG_MAX_SIZE=104857600
if not exist %LOGFILE% echo. > %LOGFILE%

:: If the log was too big, go ahead and rotate it.
del %LOGFILE%.ancient 2>NUL
rename %LOGFILE%.oldest %LOGFILE%.ancient 2>NUL
rename %LOGFILE%.older %LOGFILE%.oldest 2>NUL
rename %LOGFILE%.old %LOGFILE%.older 2>NUL
rename %LOGFILE% %LOGFILE%.old 2>NUL

java -jar pogo.scraper-all-0.3.0.jar  >> "%LOGFILE%" 2>NUL
pause


-------------------------------

u wont see any output in the cmd window, so you will need to open the log to see the google login info

pr0ves:
You can use `tee` which does the same thing as echo but prints the output on the terminal too.
It's not implemented in windows but just google `tee windows` and you'll find the version for windows

iamthasasquatch:
Using this you lose the colors in the console window however it works.

http://www.softpedia.com/get/System/System-Miscellaneous/Wintee.shtml

put that in same folder and run make a batch file that does

gradlew run | wtee logname.txt

apottere:
For linux users, you can use `tee -a out.txt` to append to the log file and not delete it every time.

iamthasasquatch:
here is the command I use for logging it to a timestamp file.

bat file named start.bat

gradlew run | wtee %time:~0,2%-%time:~3,2%-%time:~6,2%_%date:~-10,2%-%date:~-7,2%-%date:~-4,4%Log.txt

or
[...]

PR #894, opened 2016-08-09
updated logging functionality to use slf4j and logback

rraumberger:
**Fixed issue:**
#250 
#91 
#427 

**Changes made:**
1. Updated Log.kt so SLF4J Loggers are used
2. Added MarkerFilter.kt, which is a needed as a "clean" way to preserve ANSI coloring
3. Added logback.xml, including ANSI coloring and rolling-file logging
4. added default-created logs/ directory to gitignore
5. replaced "build-dir referenced as source-dir" with "source-dir referenced as source-dir" in build.gradle (version.properties is still embedded correctly in the distributable)


jabbink:
:+1: will look into this

jabbink:
Thanks!

PR #907, opened 2016-08-09
Auto-restart capability for repeating bot-loop exceptions

rraumberger:
**Fixed issue:**
Provides a possible workaround for most "my bot crashed after 2h for no reason" issues.

**Changes made:**
1. Added exception-handler to Bot.kt
2. BotService now registers an exception handler for all submitted bots
3. BotService now restarts all bots which exceed the configured autoRestartThreshold
4. replaced println-statements with proper Log.x calls

### DIFF
--- a/config.properties.template
+++ b/config.properties.template
@@ -117,6 +117,12 @@ timer_walk_to_start_pokestop=-1
 # Set profile update timer (Default: 60)
 profile_update_timer=60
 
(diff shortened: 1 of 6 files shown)


ritsu:
When the bot restarts, does it go back to the starting lat/lon? Because if the last position was, say, 3km away from the starting position, then teleporting back to the starting position might result in a soft ban.

Maybe add a delay based on distance from starting position?

PR #922, opened 2016-08-09
added feature to transfer pokemon after maxCP ratio

pashorizer:
Added option to transfer pokemon with CpMax ratio
looks like this

`Going to transfer NIDORAN_FEMALE with CP 327 and IV 62%; reason: CP < 1500 and IV < 89% and CP max 704: achieved 46% <= 60%`

One use case would be like this:

```
transfer_cp_threshold=-1
transfer_iv_threshold=89 // whatever you like as IV
transfer_cp_min_threshold=60 
```
If that Nidoran had 600CP it would result in achieved CP% ~80% and would pass the filter. So no need to use CP anymore since with that you would always filter out a Magikarp which can have max ~260CP anyways if you set `transfer_cp_threshold` above 260. Which most of us did.

OR if you want to have the very best
```
transfer_cp_threshold=-1
transfer_iv_threshold=-1
transfer_cp_min_threshold=90
```
With that setting every Pokemon you catch must have almost a perfect IV and almost max possible CP for the current trainer lvl.

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/util/pokemon/PokemonData.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/util/pokemon/PokemonData.kt
@@ -22,6 +22,11 @@ fun PokemonData.getIvPercentage(): Int {
     return ivPercentage
 }
 
+fun PokemonData.getCpPercentageToPlayer(playerlvel: Int): Int {
+    // TODO replace this when api has implemented this see Pokemon.kt
+    return -1
+}
+
 fun PokemonData.getStatsFormatted(): String {
     val details = "Stamina: $individualStamina | Attack: $individualAttack | Defense: $individualDefense"
     return details + " | IV: ${getIv()} (${(getIvPercentage())}%)"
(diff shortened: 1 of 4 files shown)


yscphdk:
How do I add it?

pashorizer:
@yscphdk Checkout on my branch

src/main/kotlin/ink/abb/pogo/scraper/util/pokemon/PokemonData.kt
```diff
@@ -22,6 +22,11 @@ fun PokemonData.getIvPercentage(): Int {
     return ivPercentage
 }
 
+fun PokemonData.getCpPercentageToPlayer(playerlvel: Int): Int {
```
jabbink: `playerlevel`

jabbink:
Would make sorting on this make sense?

andredp:
@jabbink, I think sorting by max CP already does the same sorting as %maxCP... Because if a pokemon has more %maxCP that another from the same species it will have higher CP, right? Or am I missing something?

andredp:
I was answering to @jabbink 's question. I agree with you that among different pokemons, %maxCP is a way better measurement of how good the pokemon is, but the sorting feature is only used between pokemons of the same species (when you reach the maximum amount of that pokemon, it decides which one will be transferred, in which case looking at the CP alone is enough.)

pashorizer:
Dat update button

PR #952, opened 2016-08-10
REST API Improvement

Peyphour:
I added a few features to the REST API

**Changes made:**

* Pokemon : list/transfer/evolve/power-up/favorite/show candies
* Items : list/drop/use lucky egg or incense
* Change bot location

This is not finished yet, I need at least to implement pokemon renaming, settings modification as well as authentication.

This is a discussion thread, if you want me to add other features tell me


src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
```diff
@@ -56,4 +70,160 @@ class BotController {
...
+        val data = service.getBotContext(name).api.inventories.pokebank.pokemons
+        val returnData = mutableListOf<PokemonData>()
+        for(pokemon in data) {
+            returnData.add(PokemonData().buildFromPokemon(pokemon))
+        }
+
+        return returnData
+    }
+
+    @RequestMapping("/bot/{name}/pokemon/{id}/{method}")
```
devinsba: This doesn't particularly follow rest thinking. I would recommend:

```
PUT /bot/{name}/pokemon/{id}/action

>> with a payload like:
{"action":"transfer"}
```

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
```diff
@@ -56,4 +70,160 @@ class BotController {
...
+            "favorite" -> {
+                result = pokemon!!.setFavoritePokemon(!pokemon.isFavorite).toString()
+                when(pokemon.isFavorite) {
+                    true -> Log.magenta("REST API : pokemon " + pokemon.pokemonId.name + "with stats (" + pokemon.getStatsFormatted() + " CP : " + pokemon.cp + ") is favorited")
+                    false -> Log.magenta("REST API : pokemon " + pokemon.pokemonId.name + "with stats (" + pokemon.getStatsFormatted() + " CP : " + pokemon.cp + ") is now unfavorited")
+                }
+            }
+
+            "candies" -> {
+                result = ""+pokemon!!.candy
```
jabbink: `.toString()`?

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
Peyphour: Yes I know, I made this to test more practically. Your suggestion or a different end point for each action

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
```diff
@@ -56,4 +70,160 @@ class BotController {
...
+            }
+
+            else -> {
+                result = "Unknown action (transfer/evolve/powerup/candies/favorite"
```
jabbink: `)`

jabbink:
Please don't forget to use default Kotlin format (auto format code in IntelliJ)

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
Peyphour: Yep knew this one

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
Peyphour: Until I find a better way to provide feedback

Peyphour:
Will do !

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
```diff
@@ -56,4 +69,202 @@ class BotController {
...
+
+        // Stop walking
+        ctx.pauseWalking.set(true)
+
+        ctx.lat.set(latitude)
+        ctx.lng.set(longitude)
+
+        ctx.pauseWalking.set(false)
+
+        ctx.server.setLocation(latitude, longitude)
```
jabbink: When the bot is currently walking this will immediately reset your location back to the next value on the path.

Peyphour:
Rest methods are better implemented and pokemon renaming is done. 
I noticed a bug when changing the bot location, if I call : 
`http://127.0.0.1:8080/api/bot/xxxx/location/48.648857/-2.026161`

I get this : `Getting map of (48.648892355253146, -2.000056914809637)`

Anyone know from where this come from ?

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
Peyphour: But I told the bot to stop walking, is this not enough ? Should I wait for the end of the walking task ?

Peyphour:
I'm thinking about not doing setting modification, it would too much complicated when you can simply reload the bot

jabbink:
>I get this : Getting map of (48.648892355253146, -2.000056914809637)

The bot always adds a random deviation to coords so it never sends the same location twice (instant bot/gps faker giveaway)

(stems from the origins of this bot for Ingress, which behaved humanely; and I copied the location randomization func right over)

svombergen:
will this also add possibility to check wether luckyegg or incense is currently activated?

Peyphour:
@svombergen I think no, I don't see a way to check this in the API. But if you try to use an incense or lucky egg while it is still active you should get an error

jabbink:
Can confirm: you get an error when it's activated already and that should be caught.

jabbink:
(in the evolve pr. Sorry)

Peyphour:
Authentication works now, each bot has his own auth token generated on boot. The token must be passed in the X-PGB-ACCESS-TOKEN

gise88:
In order to get the same result of https://github.com/jabbink/PokemonGoBot/pull/968 I ask if you can add to the your class `PokemonData` the following info:

- individualStamina
- individualAttack
- individualDefense
- candy
- candies to evolve
- level
- the name of move 1
- the name of move 2
- nickname
- deployedFortId (or just a boolean that tells me if the pk is in a gym)
- stamina
- max stamina
- maxCp
- getAbsoluteMaxCp()

And please add a way to get from REST API all the informations of the profile:
- name
- level
- exp
- exp to next level
- stardast
- team
- pokebank used
- pokebank limit
- backpack used
- backpack limit

Peyphour:
@gise88, i'm doing it right now, and pokedex is accessible with the current API :)

FrancYescO:
@Peyphour are you going to implement all in the rocket theme or just working for the backend?

Peyphour:
I'm just working for the backend atm, if nobody does the rocket theme I could do it

FrancYescO:
To uncomment: eggObj.distanceWalked = egg.eggKmWalked

Peyphour:
Can you create an issue for this ?

FrancYescO:
Its working, just uncomment :P

Peyphour:
I know it's working but it's better for tracking if you create an issue

FrancYescO:
https://github.com/jabbink/PokemonGoBot/issues/989

gise88:
Or just create a new rest api for the eggs: list/km-walked

Peyphour:
Will do eggs endpoint later

Peyphour:
This is finished, if you have more suggestions i'm ready to listen

REST API Documentation : 

* Secure your API with the `rest_api_password` setting. This password is shared across all bots
* Request an access token with a `POST` to `/api/bot/{name}/auth`, the request body must be the `rest_api_password`. This token must be placed in the X-PGB-ACCESS-TOKEN for further use of the API.
* Pokemon end-point : 
  - GET `/bot/{name}/pokemons` => List all pokemons
  - POST `/bot/{name}/pokemon/{id}/transfer` => Transfer pokemon
  - POST `/bot/{name}/pokemon/{id}/evolve` => Evolve pokemon
  - POST `/bot/{name}/pokemon/{id}/powerup` => Power-up pokemon
  - POST `/bot/{name}/pokemon/{id}/favorite` => Toggle favorite for this pokemon
  - POST `/bot/{name}/pokemon/{id}/rename` => Rename pokemon, request body MUST be the new name of the pokemon

* Item end-point : 
  - GET `/bot/{name}/items` => List all items
  - DELETE `/bot/{name}/item/{id}/drop/{quantity}` => Drop `quantity` of this item
  - POST `/bot/{name}/useIncense` => Use an incense
  - POST `/bot/{name}/useLuckyEgg` => Use a lucky egg

* Misc end-points : 
  - POST `/bot/{name}/location/{latitude}/{longitude}` => Change bot location
   - GET `/bot/{name}/profile` => Get account profile
  - GET `/bot/{name}/pokedex` => Get account pokedex
  - GET `/bot/{name}/eggs` => Get eggs

gise88:
Is the following a correct request?

`curl -X POST -d "ultra_secure_password" http://localhost:8001/api/bot/<PLAYER NAME>/auth`

I think I missed something! :sweat_smile:

Peyphour:
Nope, you issued the request to the GUI server. 
You should have `curl -X POST -d "ultra_secure_password" http://localhost:8080/api/bot/<PLAYER NAME>/auth` if your server port (the application one the websocket one) is 8080

gise88:
Great! Now it answers me! But....

`curl -X POST -d "ultra_secure_password" http://localhost:8080/api/bot/gitest2688/auth`

`{"timestamp":1470929146939,"status":500,"error":"Internal Server Error","exception":"java.lang.IllegalArgumentException","message":"Request processing failed; nested exception is java.lang.IllegalArgumentException: Bot gitest2688 doesn't exists !","path":"/api/bot/gitest2688/auth"}`

playername

(after your reply I will remove my username.. :smile:)

Peyphour:
You got it wrong ! This is the bot name : aka `default` if you use config.properties or the name before .json if you use .json config files

gise88:
Oooooh! Ok, now I understand.. 

And what about to another api that provides all the bot name currently running?

Peyphour:
When you call `/api/bots` you get all bots currently running. But the name property isn't included, @jabbink can we include it again ?

gise88:
There is a password leak... :smile: 
When you call the `/api/bots` it will reply with the `rest_api_password` too..
"restApiPassword":"ultra_secure_password"

However I don't get why using the following query i get `Unauthorized`:
`curl -X POST -d "ultra_secure_password" http://localhost:8080/api/bot/default/auth`

I am currently using the `config.properties` file

Peyphour:
Because the server receive `ultra_secure_password=`, you need to send a raw payload. With cURL it is not trivial, i recommend that you get this chrome app : https://chrome.google.com/webstore/detail/advanced-rest-client/hgmloofddffdnphfgcellkdfbfbjeloo?utm_source=chrome-app-launcher-info-dialog

Peyphour:
I removed the location feature until I resolve : 
 * Delta between desired location and actual location the bot is taken to
 * When teleporting, the bot will still try to walk to the location he was previously targeting

FrancYescO:
Have you looked on how is actually managed with socketserver (right click on the rocket theme map to set the new position)? Probably you already have the solution.

Ps. You could also add a get position method and with the egg things added the actual socketserver is fully replaceable (pooling is not better but just to say we can do it :P)

Peyphour:
I will look into it tomorrow or Sunday, I know the problem but don't really have time to look into it right now.

As for pooling, this isn't so bad since we're on a local connection :)

FrancYescO:
@Peyphour could you merge develop to allow jabbink to accept the pr? I really want to have it in the official code

Peyphour:
Done

Sieberkev:
@Peyphour 

* Can you set the licence header in the new files (that's probably somewhat important)?
* Personally I get a nagging feeling about a hardcoded password ("ultra_secure_password") since A LOT of users don't even take a minute to read the settings but just c/p it and never look at it again... Maybe keep it empty in the template but with a note to fill it in, and if it is empty generate it with random chars and save it in the configuration (you can search how the token value is saved)?
* A separate API.md file with your up to date [documentation](https://github.com/jabbink/PokemonGoBot/pull/952#issuecomment-239169946) would be nice :)
* Is there an endpoint to request all botnames (so one could build in pagination in a GUI for easy access to each bot)?

Impressive work tho so far dude :D

Peyphour:
Will do it tonight :) With this PR, bot names can be queried with a call to /bots (return bot settings including names).
I agree on the hard coded password

FrancYescO:
Another request:
setConfig

Peyphour:
I already looked into a 'setConfig' method and it turned out to be harder to implement, which config fields would you like to be changed on the fly ? Is it really worth it to change a config field and risk an unplanned behaviour ? And debugging this functionnality would be a pain in the ass ^^
If users really want this though I'll do it

FrancYescO:
I was thinking about adjusting speed or catch probability, Pokémon lists... fixing unwanted behaviour can be done later, just change the readed value (or simply save to config file) and warn that a reboot is better after the set

Peyphour:
@Sieberkev done
@FrancYescO isn't more easy to just change the config file if you're going to reboot anyway in the end ? Just trying to see a practical usage here

FrancYescO:
Sure, but if all can be accesible from the gui is not better, so the bot will just be a deamon

Peyphour:
I agree with you on that but it seems too much to reboot for each change in the settings. If I can find a way to change without reboot I will do it for sure :)

jabbink:
All done?

Peyphour:
Ready to merge but two functionalities missing yet : change bot location and change settings on the fly.
As they are the most difficult to implement they can come in a further version

FrancYescO:
Not enought to replace the values readed in Settings.kt ?

Peyphour:
Haven't tested yet, but I don't think it would work correctly without at least reloading the bot

PR #967, opened 2016-08-10
Automatically power up your highest IV pokemon

devinsba:
Adds a task to the profile loop to power up your highest pokemon (that meet an IV threshold)


gise88:
I have a pk with 100% iv and bad 2° skill and another one with 98% iv and good 2° skill.. I would not use this function because i prefer the second one.. 

Would not it be much better if we can just select which one power up/fully power up/evolve/transfer from the GUI?

devinsba:
While I agree, thats for the GUI programmers to handle which I am not. This get me closer to what I want.

Ideally there would be a "Pokemon Power" sorter so that we could find the best potential for someones use case. For me though this will largely suffice

PR #971, opened 2016-08-10
Adding Snipe Task and Context

KyleBoyer:
Snipe task is finished, however it is not called yet. A socket listener or something else needs to be implemented to call the SnipePokemon task.


KyleBoyer:
Closed for https://github.com/jabbink/PokemonGoBot/pull/1036

PR #1006, opened 2016-08-11
Add "next level in XY hours" to profile info

Turtizzle:
**Changes made:**

* Add "next level in XY hours" to profile info
* Added timestamp to every profile-log-message (there were 2 new lines without timestamp, this was bothering me quite a bit)


Peyphour:
Nice one

Issue #977, opened 2016-08-11
Request - GPX Build in


NoobWithBoobs:
Hello,
Requesting for GPX following to be build in to json/config file. Keep the box from traveling off course.

Do-able?

Peyphour:
Would be interesting

NoobWithBoobs:
and gpx are pretty easy to make with google earth and bunch of other programs too.

Peyphour:
Support for GPX was already in the previous version of the GUI so it shouldn't be too hard to implement

NoobWithBoobs:
I have the bot reload every 30mins due and usually i'm not around to resubmit via the gui again and again. If gpx support was build into config/json file, bot could roam around in more focus areas.

Peyphour:
Just checked, it's do-able

NoobWithBoobs:
The excitement rages!!!

staurakis:
It would be a great add on!!! :)

Peyphour:
Will do it today or tomorrow

NoobWithBoobs:
i look forward to it, keep us posted here 👍

Peyphour:
Will have two settings I think : 
* GPX file
* Number of time the file must be repeated

NoobWithBoobs:
Repeat is an awesome idea, please be sure to include -1 option as well.

Peyphour:
Of course ^^

smileytechguy:
How can I make this kind of GPX?

NoobWithBoobs:
Easist ways to make GPX files online
http://www.gpsies.com/createTrack.do?language=en

NoobWithBoobs:
@Peyphour Direction might be cool in future versions.

Peyphour:
Directions ?

NoobWithBoobs:
Direction as in backwards, forward, bidirectional.

staurakis:
I think the only logical route to have is a circular one... There is no point to directly go back the opposite route as nothing would have been spawn yet.
On the other hand, a delay time maybe a good idea if you have a small route ;)

NoobWithBoobs:
@staurakis, you do have a point with circular routes but I was thinking long roads and such. Anyways, I'm excited for just the basic, I could make a loop myself. 

Delay Time to wait for Pokemon to spawn?

staurakis:
Me too! ;)

Peyphour:
#1008

NoobWithBoobs:
[...]

PR #1008, opened 2016-08-11
Add GPX support

Peyphour:
**Fixed issue:** #977 

**Changes made:**

* Added two settings : 
  - gpxFile ; self explanatory
  - gpxRepeat : how many time does the file must be repeated by himself (0 or -1 does nothing)
* Added ReadGpx task


Peyphour:
Not ready to merge yet

src/main/kotlin/ink/abb/pogo/scraper/tasks/ReadGpx.kt
```diff
@@ -0,0 +1,47 @@
...
+
+        val coords = document.rootElement.getChild("trk", document.rootElement.namespace)
+                        .getChild("trkseg", document.rootElement.namespace)
+
+        var i: Int = settings.gpxRepeat
+
+        Log.green("GPX plugin : reading file " + settings.gpxFile + " with $i repeats")
+
+        // gpxRepeat == 0 or -1 : No coordinates will be added
+        while(i > 0) {
```
jabbink: Could you maybe put this in a task to add the next iteration when the previous ended. That way an "infinite" option is possible in a "clean" way (instead of adding your route eg "99999" times, which can be very memory intensive)

Peyphour:
Still needed ?

PR #1022, opened 2016-08-12
Add Poppo theme to UI

pureexe:
I have create UI theme for PokemonGoBot   
Thank for create great Bot!

**Preview**

"More maps less data"  

"Easy Access from right conner"  

"Keep data floating over map"


pureexe:
@smileytechguy Oh! thank you for your suggestion i will add menu hide console to settings and limit log as 1000 line for prevent browser from crash

Issue #1047, opened 2016-08-13
[Feature Request] Randomize Bot Behaviour


AlexBoulyga:
Dear PokemonGoBot Developers,
I looked through your code which I found really interesting, and noticed it missed some simple randomized behaviour.
In my mind I have: Curved Rate instead of true or false set to a value (so that its sometimes curved sometimes not).
Randomized speed in a definable range around the normal speed property
Random stops and pauses for a random time, both settable (wait chance, minimal wait time, maximal wait time)

I have already wrote some working code for that, and would like to contribute it to the project if thats ok.

greetings and thanks for the great bot

jabbink:
Nice. Another feature could be to sometimes miss pokeballs. And wait between throws and evolves.

shilch:
Yes, I think this feature is necessary. https://www.reddit.com/r/pokemongodev/comments/4xct8f/psa_permabans_have_arrived_stop_all_use_of_3rd/

AlexBoulyga:
Thats what I also thought, but I hadnt enough time to look Into it yet

AlexBoulyga:
I would even say not only miss, as its easy to Hut the Pokémon, but to throw different kinds of balls linke nice, great etc.

AlexBoulyga:
Ok I analyzed the code, there is already this catching probability for desired/undesired catches. I logged the number of tries, and sometimes the bot needs one, sometimes even 3 tries.
After further digging I found out that this probability only affects which kind of ball you use and similar parameters.
As far as I can say now it always hits the same spot (which is really obvious for niantic/google) which should be changed and the size of the coloured circle is always nearly the same.
I managed to randomize it a little, however it needs better testing and beeter/cleaner code, so it might take a day or even two (considering its weekend now). Anyways, I'll keep this updated

gise88:
Setting.tk (commit https://github.com/jabbink/PokemonGoBot/commit/983ae4ec2fb802394b7c7a32cb63fd4682ca5e52) is still read option "always_curve" (https://github.com/jabbink/PokemonGoBot/blob/develop/src/main/kotlin/ink/abb/pogo/scraper/Settings.kt#L94) but in the config.properties.template (https://github.com/jabbink/PokemonGoBot/blob/983ae4ec2fb802394b7c7a32cb63fd4682ca5e52/config.properties.template#L120) the option "always_curve" was replaced with "desired_curve_rate".

AlexBoulyga:
[...]

PR #1054, opened 2016-08-13
Bugfix, wrong settings value was used

AlexBoulyga:
#1047 

Fixed a bug @gise88 pointed out, the old vsettings property was used.


gise88:
I don't know if it is a bug but in the other options there aren't the spaces before and after the '='.. :grin: 
https://github.com/jabbink/PokemonGoBot/blob/develop/config.properties.template#L120

AlexBoulyga:
Bug no but style guideline (facepalm). I shouldnt do any Bugfixes when Im in a hurry

PR #1063, opened 2016-08-14
Update USAGE.md

ruchern:
Typo error in the instruction set.


Sieberkev:
This is not a typo but intentional, there's a gradlew script in the root of the project to make sure it does not use your local version.

ruchern:
@Sieberkev Yes. Sorry. I realised that immediately after I edited that.

PR #1075, opened 2016-08-14
Check LevelUp rewards & Badges

Sieberkev:
**Fixed issue:** Check if there are level and/or badges rewards to claim

**Changes made:**

* Check LevelUpRewards & Badges

This needs some reviewing and testing :smile:


nhodges:
Is this a UI-only issue? The rewards aren't automatically distributed as you level up? (Does the bot miss all the lures/incubators as it levels up?)

Sieberkev:
> Does the bot miss all the lures/incubators as it levels up?
At this moment, it does... This PR is to fix that :)

jabbink:
Is there a way to request until which level these were requested already?

If not, how does the app request these upgrades? Seems like a dead-giveaway for botting that we request this every time we log in.

Sieberkev:
I did not find such a request in the API :(

AlexBoulyga:
I habe Götzen for my bot lvl 18 all items since lvl 2 after starting the bot implemeting the code from Here.

PR #1076, opened 2016-08-14
Throttle evolving/transfers to avoid "null" exceptions on large numbers

Sieberkev:
**Fixed issue:** Evolving/transferring a large amount of pokes (50+) throws "null" exception

**Changes made:**

* Throttle after evolve/transfer to avoid "null" exceptions on very large (100) evolves

This needs testing!


src/main/kotlin/ink/abb/pogo/scraper/tasks/EvolvePokemon.kt
```diff
@@ -67,9 +67,11 @@ class EvolvePokemon : Task {
                 if (settings.evolveBeforeTransfer.contains(it.pokemonId)) {
                     Log.yellow("Evolving ${it.pokemonId.name} before transfer")
                     val evolveResult = it.evolve()
```
sbeckeriv: I gave this a whirl. I had to add a timeout before the it.evolve() call. Seemed to work fine then.

Issue #1040, opened 2016-08-13
[Feature Request] - Bot Timeout & Stack evolves without lucky egg


gise88:
## Feature requests/proposals
I have two requests:

- Sometimes, before going to bed or going out, I turn on the bot. It would be possible to put a kind of time-out to make sure that the bot will stops after a certain time? I don't want to run the bot so much time continuously.. 

- I really do like the last feature "stack-evolves" but I don't like the fact that it will use my lucky eggs.. Can we choose if the bot has to use them or not? To me the bot have to be a free help.. When there will be no bot (hopefully not soon) these eggs will be very important!

Thank you for your amazing job!

badeip:
On linux it should be possible to use `timeout` for the former request (untested)

Run from 1 to 3 hours:

    timeout $(($RANDOM % (10800 + 1 - 3600) + 3600)) java -jar ./PokemonGoBot-0.5.2.jar

gise88:
@badeip Thanks for this but some of my friends are Windows users.. I am a dev but is my first time with Kotlin! :grin: If someone expert want to implement those feature I believe that the whole community will enjoy it.. :smile:

Sieberkev:
First one has been implemented by @1095
Second one has been implemented by @1082

PR #1082, opened 2016-08-15
Added variable use_lucky_egg for evolving

svombergen:
Added variable use_lucky_egg (1 = use, anything else = do not use)

As requested in
#1040


config.properties.template
```diff
@@ -164,6 +164,9 @@ evolve_before_transfer=CATERPIE,RATTATA,WEEDLE,PIDGEY
 # The stack of evolves needed to pop lucky egg and evolve all
 evolve_stack_limit=100
 
+# Use lucky egg before transfer (1 = yes, anything else = no)
```
jabbink: `Use lucky egg before evolve & transfer` to avoid confusion I guess

PR #1095, opened 2016-08-15
Added optional delays between transfer/evolve/drop and bot stop after reaching settable limits

AlexBoulyga:
Issue #1047
#1078 
#1064

**Changes made:**
Added optional delays between transfer/evolve/drop and optional bot timeouts, which now successfully terminate the application
Added option to never catch certain pokemon

PR #1109, opened 2016-08-16
Feature: Thresholds

muschter:
This pull request adds thresholds for a maximum amount of looted Pokestops and caught Pokemon. This will prevent the bot from continuously catching Pokemon and looting Pokestops, which normally leads to your account being flagged.

**Changes made:**

* Added a threshold for caught Pokemon and looted Pokestops
* Delay the tasks for 12 hours before resetting the threshold.

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
@@ -131,7 +131,11 @@ class SettingsParser(val properties: Properties) {
 
                 waitTimeMin = getPropertyIfSet("Minimal time to wait", "wait_time_min", defaults.waitTimeMin, String::toInt),
 
-                waitTimeMax = getPropertyIfSet("Maximal time to wait", "wait_time_max", defaults.waitTimeMax, String::toInt)
+                waitTimeMax = getPropertyIfSet("Maximal time to wait", "wait_time_max", defaults.waitTimeMax, String::toInt),
+
+                pokestopThreshold = getPropertyIfSet("The maximum amount of pokestops to loot before stopping", "pokestop_threshold", defaults.pokestopThreshold, String::toInt),
+
+                pokemonThreshold = getPropertyIfSet("The maximum amount of pokemon to catch before stopping", "pokemon_threshold", defaults.pokemonThreshold, String::toInt)
         )
     }
 
@@ -266,7 +270,10 @@ data class Settings(
 
         val waitChance: Double = 0.0,
         val waitTimeMin: Int = 0,
-        val waitTimeMax: Int = 0
+        val waitTimeMax: Int = 0,
+
+        val pokestopThreshold: Int = 1500,
+        val pokemonThreshold: Int = 1000
 ) {
     fun withName(name: String): Settings {
         this.name = name
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
@@ -28,6 +28,13 @@ import ink.abb.pogo.scraper.util.pokemon.shouldTransfer
 class CatchOneNearbyPokemon : Task {
     override fun run(bot: Bot, ctx: Context, settings: Settings) {
         // STOP WALKING
+
+        if (ctx.caughtPokemon.get() > settings.pokemonThreshold) {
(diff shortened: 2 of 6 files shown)


muschter:
Ready for review.

AlexBoulyga:
Hmm it is somehow similar to my changes
https://github.com/jabbink/PokemonGoBot/pull/1095
However my changes simply stop the bot completely (instead stopping the single task)
Probably yours is better thought through

AlexBoulyga:
I believe yours is better, in any case i think 1000 is too much for Pokémon as i heard thats the value in ca 24h you get Banned for. Ah and do you know how i can "rewind" the last commits from my pull request here in github?

AlexBoulyga:
Ah and one more thing, could you also make the waittime which is 12h choosable from settings please?

muschter:
@AlexBoulyga My apologies, I haven't checked pull requests recently. I had thought of how it would be beneficial to continue to walk around and hatch eggs, but maybe stopping entirely would be more human-like which could lead to less flagging.

muschter:
@AlexBoulyga I don't know how to rewind commits, unfortunately. However, I will go ahead and change the values and add a setting for the delay time in the morning.

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
```diff
@@ -131,7 +131,11 @@ class SettingsParser(val properties: Properties) {
 
                 waitTimeMin = getPropertyIfSet("Minimal time to wait", "wait_time_min", defaults.waitTimeMin, String::toInt),
 
-                waitTimeMax = getPropertyIfSet("Maximal time to wait", "wait_time_max", defaults.waitTimeMax, String::toInt)
+                waitTimeMax = getPropertyIfSet("Maximal time to wait", "wait_time_max", defaults.waitTimeMax, String::toInt),
```
jabbink: *Maximum

src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt
```diff
@@ -28,6 +28,13 @@ import ink.abb.pogo.scraper.util.pokemon.shouldTransfer
 class CatchOneNearbyPokemon : Task {
     override fun run(bot: Bot, ctx: Context, settings: Settings) {
         // STOP WALKING
+
+        if (ctx.caughtPokemon.get() > settings.pokemonThreshold) {
+            Thread.sleep(3600 * 12000)
```
jabbink: instead of sleeping, maybe set the settings.catchPokemon to false and store the time it was disabled and then re-enable it later on. See how this is done with a full pokebag: https://github.com/muschter/PokemonGoBot/blob/83c9dc6077f63f34342534076101615e0a87701a/src/main/kotlin/ink/abb/pogo/scraper/tasks/CatchOneNearbyPokemon.kt#L141

src/main/kotlin/ink/abb/pogo/scraper/tasks/ProcessPokestops.kt
```diff
@@ -50,6 +50,12 @@ class ProcessPokestops(var pokestops: MutableCollection<Pokestop>) : Task {
         if (startPokestop == null)
             startPokestop = sortedPokestops.first()
 
+        if (ctx.lootedPokestops.get() > settings.pokestopThreshold) {
+            Thread.sleep(3600 * 12000)
```
jabbink: same as with catching pokemon limit: maybe just disable the setting.

src/main/kotlin/ink/abb/pogo/scraper/Settings.kt
AlexBoulyga: It was me, ill change it in my commit :)

HoreaM:
*not sure here is the best place to do this, I'm new to GitHub, sorry for my noobishness* 

Would it complicate things a lot to change it so that it keeps the values for Pokemons caught so far and Pokestops visited so far somewhere locally, so even if let's say the bot crashes or is stopped, if restarted it will continue from the previous values.

src/main/kotlin/ink/abb/pogo/scraper/tasks/ProcessPokestops.kt
muschter: @jabbink I can reassign the settings.catchPokemon value, but not the settings.lootPokestop value.

muschter:
@HoreaM I believe this has recently added in #1129

PR #1113, opened 2016-08-16
Changed the route provider

pr0ves:
Changed it back since we were destroying the poor yournavigation.com.
Keep in mind that with router.project-osrm.com the bot will follow the route made for a car, so longer routes.


Sieberkev:
Aaaaaand RIP OSRM ^-^

> Failed to load resource: the server responded with a status of 503 (Service Unavailable: Back-end server is at capacity)

PR #1116, opened 2016-08-16
Fix rest token

Peyphour:
Allow REST API token to be in each request URL.

example : `/api/bot/{name}/pokemons?token=...`


Peyphour:
Oops, created the branch from the wrong one

PR #1117, opened 2016-08-16
Hatch largest eggs in disposable incubators

KyleBoyer:
**Changes made:**

* Filtered the incubators descending by uses remaining


src/main/kotlin/ink/abb/pogo/scraper/tasks/HatchEggs.kt
```diff
@@ -38,7 +38,9 @@ class HatchEggs : Task {
         val incubators = ctx.api.cachedInventories.incubators
         val eggs = ctx.api.cachedInventories.hatchery.eggs
 
-        val freeIncubators = incubators.filter { !it.isInUse }
+        val freeIncubators = incubators
+                .filter { !it.isInUse }
+                .sortedByDescending { it.usesRemaining }
```
jabbink: just to also put this here and not only in Discord: the infinite use incubator has usesRemaining set to 0 or -1.

KyleBoyer: That is what I also discovered, therefore the sort works perfectly without any additional code! (:

PR #1118, opened 2016-08-16
REST API fix : allow CORS requests

Peyphour:
Fix CORS calls for the REST API


Sieberkev:
Nice work, works like a charm now :heart:

Issue #144, opened 2016-07-23
Reuse last location option


Tjessx:
After 20-30 minutes, I need to restart this bot, and change the lat&long manually.
It would be great if there was an option to automatically use the latest location

EDIT by @jabbink: add as config flag with default `true`, please.

Ixath:
While I do agree that it would be neat, why not just wait 1 or 2 minutes and restart the bot from the original location?

cglatot:
If you login via PTC instead of Google, you should not need to restart the bot. It's a problem with the current API release

CptJackieSparrow:
@lxath in order to have good farm you need to walk and find pokemon. Spawns take approx 7-8 minute to refresh. Also 5 minutes is the time for pokestop refresh. 1-2 mins would be short, but i'm restarting bot every 8 minutes. Never got softbanned.

pr0ves:
The problem of the token expiration is goong to be fixed in https://github.com/Grover-c13/PokeGOAPI-Java/pull/165. 
We'll still going to add a way to save last coordinates.

wartickler:
If we get Sqlite integration this is trivial.

apottere:
I think this might not be as necessary as you think - when I use the bot in the middle of my town, it finds 218 pokestops.  Assuming it takes 15 seconds to walk from one to the other, it will never run out of pokestops in the area.

Tjessx:
@apottere it would not run out of poke stops, but let's say it walks a few miles/km further, and you restart it immediately, and it jumps back those few miles/km because the coordinates are still the same as last time you ran it.
At this point it wouldn't be a big problem, but if they would improve their gps spoof security there could be a lot of soft bans

wartickler:
I've already been soft banned for jumping too far with this.

viktor89:
I agree with this enhancement. Perhaps the bot could just write to the properties-file, the same way it writes the token. Just have it update the location settings in the properties-file if some setting has been enabled. This would allow more automation in running/restarting the bot as service, etc.

gabrielklein:
+1 It would be great to save the latest position of the bot!
(Or add an extra option :) )

vamirineni87:
anyone know the status on this?

Peyphour:
Taking this issue

KyleBoyer:
> Taking ownership

I already have most of this done, but my InteliJ merging screwed stuff up

Peyphour:
I'll leave it to you then ^^

[...]

PR #1129, opened 2016-08-16
Save last location

Peyphour:
**Fixed issue:** #144 

**Changes made:**

* Update settings lat/lng on bot stop. JSON file will get automatically updated with the last location

### DIFF
--- a/config.properties.template
+++ b/config.properties.template
@@ -18,6 +18,7 @@ proxy_password=
 token=
 latitude=
 longitude=
+save_location_on_shutdown=true
 
 # Desired probability to catch Pokemon with 1 throw
 desired_catch_probability=0.4
--- a/json-template.json
+++ b/json-template.json
@@ -1,6 +1,7 @@
 {
   "latitude" : 0,
   "longitude" : 0,
+  "saveLocationOnShutdown": true,
   "credentials" : {
     "type" : "ptc",
     "username" : "",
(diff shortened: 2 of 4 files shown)


KyleBoyer:
This won't write out to the .properties file tho. The required function to be used is settings.writeProperty(fileName, "key", "value")

Peyphour:
Don't know you followed one of my last PR but now all config files are JSON files.
Which means that : 
  - If you create a .properties it will be migrated to a .json on the bot shutdown
  - If you change a setting during runtime, it will thus be saved in the .json file

AlexBoulyga:
Just a question, If you dont usw an IDE, how do you change those settings?

AlexBoulyga:
Before i noticed that i was searching for over half an hour why the bot does not use my changes made in the config.properties

Peyphour:
You don't need an ide to change to .json  config file

AlexBoulyga:
OK then thats good :)

FrancYescO:
am i wrong or this is fully useless? is saving cordinates just to a var when closing?!

probably was better implemented in https://github.com/jabbink/PokemonGoBot/pull/1130 but both have the issue https://github.com/jabbink/PokemonGoBot/issues/1142

Peyphour:
When closing, the bot's current settings are saved to a json file. 
And when there is at least one json file, the . properties isn't loaded anymore

Peyphour:
Did you read my answer to Kyle who asked the same thing ? ...

PR #1130, opened 2016-08-16
Save lat/long at bot end & better egg placement

KyleBoyer:
**Changes made:**

* Hatch furthest KM eggs in disposable incubators in order maximize their usage.
* Save lat/long in properties file at bot end


jabbink:
What does this do with json configs?

Taraegon:
Excellent move to save lat/long in config file.

One way I use to avoid a ban is to ensure I log in near my previous log out IF there is only a short time delay. This might help somewhat.

(Players can't leave it up to the bot to do everything. We have to play smart too)

FrancYescO:
https://github.com/jabbink/PokemonGoBot/pull/1129

Issue #1122, opened 2016-08-16
Send correct elevation


jabbink:
**Description:**
Send semi-correct altitude information

**Expected behavior:**
Send better altitude information.

Cache hard, only query altitude and slightly randomize it for a single S2 level 15 (or larger) cell.

Sources:
- https://mapzen.com/documentation/elevation/
- https://developers.google.com/maps/documentation/elevation/intro

Leave this sentence in your issue as proof that you have read and used this template.

KyleBoyer:
https://elevation.dev.mapzen.com/height?json={%22shape%22:[{%22lat%22:40.7829,%22lon%22:-73.9654}]}

PR #1155, opened 2016-08-17
Use correct altitude for given GPS coordinates

KyleBoyer:
Fixes: https://github.com/jabbink/PokemonGoBot/issues/1122

**Changes made:**

* Looks up correct elevation/altitude in meters for given GPS coordinates
* Fixed 2 warnings(by changing var to val)

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/Context.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/Context.kt
@@ -56,9 +56,9 @@ data class Context(
             return this.s2Cache[cellId]!! + rand
         }
         var elevation = 10.0
-        var url = HttpUrl.parse("https://maps.googleapis.com/maps/api/elevation/json?locations=$latitude,$longitude&sensor=true").newBuilder().build()
-        var request = okhttp3.Request.Builder().url(url).build()
         try {
+            val url = HttpUrl.parse("https://maps.googleapis.com/maps/api/elevation/json?locations=$latitude,$longitude&sensor=true").newBuilder().build()
+            val request = okhttp3.Request.Builder().url(url).build()
             val result: Map<*, *>
             result = ObjectMapper().readValue(OkHttpClient().newCall(request).execute().body().string(), Map::class.java)
             val results = result["results"] as List<*>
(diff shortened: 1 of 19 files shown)


Sieberkev:
Nice work with the fetching, but won't this nuke their server like hell? That's probably why @jabbink asked for caching on S2 level. May I propose a slight modification?

Check in which [S2 cell](http://blog.christianperone.com/2015/08/googles-s2-geometry-on-the-sphere-cells-and-hilbert-curve/) the requested lat/long is (the library to do so is already in the bot), fetch the altitude only once and cache/use the cell size (= level) + altitude (with some small randomisation per lat/long when they are in large cells).

Save/load the cached altitudes/sizes in a local cache file (on start/stop) so on a bot restart (or if you use multiple bots in the same area) you use the cached values (it's not that the earth changes altitude so much) ;)

This would be way more friendly to the external API.

AlexBoulyga:
Great idea , however i believe getting the altitude on each Start wont overload their servers and will be more precise. F.e. If I change the location my bot is farming, the altitude should change too. I believe the besteht way would be getting it once on start and thats was it

Sieberkev:
> however i believe getting the altitude on each Start wont overload their servers

If you start every time on the exact same location, that altitude will never change (maybe when a volcano erupts, but what are the chances of that).
If you are caching anyway (because nuking other APIs/servers for information that does not change it not very polite), you better use that value too on start since you have it anyway.

> If I change the location my bot is farming, the altitude should change too.

Each time you move, it checks in which S2 cell you are. If that value is cached, use that value (+/- a small randomisation on large cells), if you don't have it in cache, fetch it from the external server.

KyleBoyer:
@Sieberkev I have made those according changes

src/main/kotlin/ink/abb/pogo/scraper/Context.kt
```diff
@@ -35,10 +40,30 @@ data class Context(
...
-)
+) {
+    fun getAltitude(latitude: Double, longitude: Double): Double {
+        val rand = (Math.random() * 3) + 1
+        val cellId = S2CellId.fromLatLng(S2LatLng.fromDegrees(latitude, longitude)).parent(15).id().toString()
+        if (this.s2Cache.containsKey(cellId) && this.s2Cache[cellId] != null){
+            return this.s2Cache[cellId]!! + rand
+        }
+        var elevation = 10.0
+        val url = HttpUrl.parse("https://elevation.mapzen.com/height?json={\"shape\":[{\"lat\":" + latitude.toString() + ",\"lon\":" + longitude.toString() + "}]}").newBuilder().build()
```
mineme45: Using Kotlin, you can concat it like this:
"https://elevation.mapzen.com/height?json={\"shape\":[{\"lat\":$latitude,\"lon\":$longitude}]}"

src/main/kotlin/ink/abb/pogo/scraper/Context.kt
```diff
@@ -35,10 +41,41 @@ data class Context(
...
-)
+) {
+    fun getAltitude(latitude: Double, longitude: Double): Double {
+        val rand = (Math.random() * 3) + 1
+        val cellId = S2CellId.fromLatLng(S2LatLng.fromDegrees(latitude, longitude)).parent(15).id().toString()
+        if (this.s2Cache.containsKey(cellId) && this.s2Cache[cellId] != null) {
+            return this.s2Cache[cellId]!! + rand
+        }
+        var elevation = 10.0
+        var url = HttpUrl.parse("https://maps.googleapis.com/maps/api/elevation/json?locations=$latitude,$longitude&sensor=true").newBuilder().build()
```
FrancYescO: put this the try

PR #1166, opened 2016-08-18
implement setlocation in REST api

FrancYescO:
Fixing https://github.com/jabbink/PokemonGoBot/pull/952/commits/f460c73414b6ac54bee518abdeb8016997eac839

just taken the code from socketserver


Peyphour:
Thanks !

Peyphour:
But is it possible to set really the bot location and not add a location to go to ?

FrancYescO:
theoretically

ctx.pauseWalking.set(true) //hopefull it's enought to completely stop the walk
ctx.lat.set(...)
ctx.lng.set(...)
ctx.pauseWalking.set(false)

should do the teleport trick

Peyphour:
Wasn't what I did ?

FrancYescO:
lOl

so the issue is pauseWalking, it pauses the walk and does not stop it, it's a bit tricky with the actual code to do a teleport

the easies way is to export the local `remainingSteps` value of `walkPath` func end set it to 0 before `ctx.pauseWalking.set(false)`

Issue #1146, opened 2016-08-17
[Feature Request] Add different route providers


AlexBoulyga:
**Description:**
I thought about adding different route providers, choosable from settings,
so that if I see one is not working I can try another one.
By this, the load would be divided between multiple providers, increasing the overall speed

I dont know however how much work that is.
I believe @pr0ves  was the one who changed it the last time
What do you think about that?

Leave this sentence in your issue as proof that you have read and used this template.

muschter:
I think this is a good idea, so we aren't stressing one provider too much.

mineme45:
I've made a mapzen route provider.
I will PR it today if everything is ok :)

Sieberkev:
Added with ##1167

Issue #1162, opened 2016-08-18
Bug - [default: PokestopLoop] - Error running loop PokestopLoop!


DiogoRNeves:
**Description:**
After the error the bot seems to stop walking.

**Steps to reproduce:**

1. Run the bot

**Expected behavior:**
N/A

**Actual behavior:**
N/A

**Stacktrace (If it's a crash):**
18 Aug 08:09:29 [default: PokestopLoop] - Error running loop PokestopLoop!
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
	at java.util.ArrayList.rangeCheck(ArrayList.java:653)
	at java.util.ArrayList.get(ArrayList.java:429)
	at ink.abb.pogo.scraper.util.directions.RouteRequestKt.getRouteCoordinates(RouteRequest.kt:46)
	at ink.abb.pogo.scraper.util.directions.RouteRequestKt.getRouteCoordinates(RouteRequest.kt:84)
	at ink.abb.pogo.scraper.tasks.Walk.walkRoute(Walk.kt:81)
	at ink.abb.pogo.scraper.tasks.Walk.walk(Walk.kt:70)
	at ink.abb.pogo.scraper.tasks.Walk.run(Walk.kt:63)
	at ink.abb.pogo.scraper.Bot.task(Bot.kt:262)
	at ink.abb.pogo.scraper.tasks.ProcessPokestops.run(ProcessPokestops.kt:74)
	at ink.abb.pogo.scraper.Bot.task(Bot.kt:262)
	at ink.abb.pogo.scraper.Bot$start$5.invoke(Bot.kt:166)
	at ink.abb.pogo.scraper.Bot$start$5.invoke(Bot.kt:36)
	at ink.abb.pogo.scraper.Bot$runLoop$1.invoke(Bot.kt:204)
	at ink.abb.pogo.scraper.Bot$runLoop$1.invoke(Bot.kt:36)
	at kotlin.concurrent.ThreadsKt$thread$thread$1.run(Thread.kt:18)

**Version:**
f9c163e1a481952616b9ef9bb77f694cd2ad15d5

**Operating System:**
Mac OS X

**Java version:**
java version "1.8.0_91"
Java(TM) SE Runtime Environment (build 1.8.0_91-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.91-b14, mixed mode)

Leave this sentence in your issue as proof that you have read and used this template.

Sieberkev:
[...]

PR #1167, opened 2016-08-18
Add Mapzen route provider, choosable from settings

mineme45:
**Feature Request:** #1146

**Changes made:**

* New Route Provider : Mapzen (valhalla.mapzen.com)
* Choose Route Provider using settings
* Support 4 Routes Providers : MAPZEN, MOBROUTING, PROJECTOSM, YOURNAVIGATION

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteProviderEnum.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteProviderEnum.kt
@@ -1,0 +1,134 @@
+package ink.abb.pogo.scraper.util.directions
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.pokegoapi.google.common.geometry.S1Angle
+import com.pokegoapi.google.common.geometry.S2LatLng
+import java.util.*
+import java.util.regex.Pattern
+
+
+enum class RouteProviderEnum {
+
+    MAPZEN {
+        override fun createURLString(startLat: Double, startLong: Double, endLat: Double, endLong: Double): String {
+            return "http://valhalla.mapzen.com/route?json={\"locations\":[{\"lat\":$startLat,\"lon\":$startLong},{\"lat\":$endLat,\"lon\":$endLong}],\"costing\":\"pedestrian\",\"directions_options\":{\"narrative\":\"false\"}}"
+        }
+
+        override fun parseRouteResponse(routeParsed: String): ArrayList<S2LatLng> {
+
+            val jsonRoot = ObjectMapper().readTree(routeParsed)
+            val status = jsonRoot.path("trip").path("status").asInt()
+
+            // status 0 == no problem
+            if (status == 0) {
+                val shape = jsonRoot.path("trip").findValue("shape").textValue()
+
+                // Decode the route shape, look at https://mapzen.com/documentation/turn-by-turn/decoding/
+                val precision: Double = 1E6
+                val latlngList = ArrayList<S2LatLng>()
+                var index: Int = 0
+                var lat: Int = 0
+                var lng: Int = 0
+                while (index < shape.length) {
+                    var b: Int
+                    var shift = 0
+                    var result = 0
+                    do {
+                        b = shape[index++].toInt() - 63
+                        result = result or (b and 0x1f shl shift)
+                        shift += 5
+                    } while (b >= 0x20)
+                    val endLat = if (result and 1 != 0) (result shr 1).inv() else result shr 1
+                    lat += endLat
+
+                    shift = 0
+                    result = 0
+                    do {
+                        b = shape[index++].toInt() - 63
+                        result = result or (b and 0x1f shl shift)
+                        shift += 5
+                    } while (b >= 0x20)
+                    val endLong = if (result and 1 != 0) (result shr 1).inv() else result shr 1
+                    lng += endLong
+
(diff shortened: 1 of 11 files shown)


Sieberkev:
Nice and clean!

Minor detail, but could it be possible to merge the keys `follow_streets` and `follow_streets_provider` (and alter the description a bit)? So if `follow_street` is empty (or the string false so old configurations don't break) = don't follow routes, or fill in one of your routeproviders in that key to use that provider (= same behaviour as the `export` config). It's just to avoid adding too many configuration keys to the already heavy configuration ^_^

I see that the new system doesn't set a `User-Agent` (or other "human like") header(s), wouldn't that give it away a bit too fast that a JAVA bot (instead of a normal browser) is making the connections?

pr0ves:
Mapzen limits the requests without the api key to 6 every minute.
Some test would be needed to check if it's enough in every case.

mineme45:
> about merging follow_streets and follow_streets_provider
It's a great idea! I will look how I can make it since I use Enum.valueOf to load settings

> about User-Agent
We call only API, so I think they know it's not a human ^^'
But I can put one it does not cost much :)

Sieberkev:
> We call only API, so I think they know it's not a human ^^

I log all User-Agent headers (and a bunch of other stuff) on our API service to see if they don't share their keys over multiple clients :p They can do something like that too ;)

src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteProviderEnum.kt
```diff
@@ -0,0 +1,134 @@
...
+    },
+
+    MOBROUTING {
+        override fun createURLString(startLat: Double, startLong: Double, endLat: Double, endLong: Double): String {
+            return "http://mobrouting.com/api/dev/gosmore.php?flat=$startLat&flon=$startLong&tlat=$endLat&tlon=$endLong&v=foot&fast=1&layer=mapnik"
+        }
+
+        override fun parseRouteResponse(routeParsed: String): ArrayList<S2LatLng> {
+            if (!routeParsed.contains("<distance>0</distance>")) {
+                val matcher = Pattern.compile("(|-)\\d+.\\d+,(|-)\\d+.\\d+").matcher(routeParsed.split("<coordinates>")[1])
```
Sieberkev: If this external service returns a 522 error (because overload), the routeParsed.split()[1] part will throw an exception and lock the walking completely (see also #1162).

src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteProviderEnum.kt
```diff
@@ -0,0 +1,134 @@
+package ink.abb.pogo.scraper.util.directions
```
Sieberkev: Could you add the licence header here?

src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteRequest.kt
```diff
@@ -1,85 +1,34 @@
 package ink.abb.pogo.scraper.util.directions
```
Sieberkev: Could you add the licence header here?

src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteProviderEnum.kt
mineme45: :)

src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteProviderEnum.kt
mineme45: Done

src/main/kotlin/ink/abb/pogo/scraper/util/directions/RouteRequest.kt
mineme45: Done

gise88:
Works great! However i don't understand why we have to choose a provider.. It may not be the bot that test which one works and from then on only using this as long as receive a response? 
On the contrary, if it receives a bed answer then it can switch to the others provider choosing by himself which one is better (ms to get the response?)..

mineme45:
@gise88 
I will mix your idea with @Sieberkev idea:

For example:
follow_streets=TOTO -> unknow provider, do not use route provider

Another example:
follow_streets=MAPZEN,MOBROUTING,GOOGLE (If we add @pr0ves google implementation in the Enum)
We will try MAPZEN, if bad response, (ban it for X min), we will try the next one: MOBROUTING, if bad response (ban it for X min), we will try GOOGLE, if bad response  (ban it for X min) don't use route provider.

If MOBROUTING working, the next time we will don't use MAPZEN, because banned for X min, and use MOBROUTING (2nd position on the user settings).

With this, user still be able to choose route provider.

Another idea for MAPZEN, add the mapzen api key on settings, if not set, call the api without api_key (limited to 6 call/min).

Sieberkev:
Afaik you need a key for Google too (for more requests?), so that could also be included in the settings?

mineme45:
Yes. I think its done by @pr0ves.

pr0ves:
I'll make a pr with the google implementation after that this and #1155 are merged.
Also @mineme45 you could add a setting field for the mapzen api key, so that the user could add one of his and avoid the query error.

Sieberkev:
FYI #1155 just got merged (= google elevation + mapzen elevation **without** API keys)

PR #1185, opened 2016-08-20
Fix YourNavigation + Project-OSM, working again + pokestop not on the street fix

mineme45:
**Changes made:**

* Fix routes providers: YOURNAVIGATION and PROJECTOSM, working again
* If we follow streets but when reached destination, the pokestop is too far, go directly


mineme45:
toto

Issue #1125, opened 2016-08-16
Add forced checkstyle test


jabbink:
**Description:**
Too many PR's add code that does not follow the default indentation and adding commits every once in a while that just re-indent code (and break other PR's) sucks.

Leave this sentence in your issue as proof that you have read and used this template.

AlexBoulyga:
Good idea (i know i am one of the culprits, but a test that does this would make my life easier, and others respectively)

mineme45:
The goal is to check this with travis?

Sieberkev:
Merged with #1246 

More checks can be added if you want! ;)

PR #1189, opened 2016-08-20
Add checkstyle to Gradle build

Sieberkev:
**Fixed issue:** #1125

**Changes made:**

* Added checkstyle to build.gradle (only check "src/main/kotlin" and not the libraries)
* Added a checkstyle.xml I found in another JAVA/kotlin project
* Fixed several checkstyle errors (File contains tab characters, Space needed before opening parenthesis, Line has trailing spaces) so the current checkstyle.xml builds without errors

**To dos**

- [x] Make checkstyle work on `gradlew build`
- [ ] Implement the default IntelliJ checks
- [ ] Why does the unused imports/star imports check doesn't work?
- [ ] Imports in lexographical order
- [ ] Make sure all current files are corrected


checkstyle.xml
```diff
@@ -0,0 +1,150 @@
...
+    <property name="message" value="Space needed before opening parenthesis."/>
+  </module>
+
+  <!-- For each spacing -->
+  <module name="RegexpSingleline">
+    <property name="format" value="^\s*for \(.*?([^ ]:|:[^ ])"/>
+    <property name="message" value="Space needed around ':' character."/>
+  </module>
+
+  <module name="TreeWalker">
```
Sieberkev: For some reason not a single TreeWalker submodule check works but there's no error whatsoever.

And my googlefu is not helping me :cry:

Zizekftw:
TreeWalker submodules have been breaking as you advance JDK versions. At work, we've had to remove those checks from the XML file. DoubleCheckedLocking and RedundantThrows used to work with Java 6, but haven't worked with 7 or 8.

http://checkstyle.sourceforge.net/releasenotes.html#Release_6.2 (For removal of Redundant Throws)
http://stackoverflow.com/questions/12795935/checkstyle-eclipse-plugin-doesnt-work (For removal of DoubleCheckedLocking).

Checkstyle has a history of making non-backwards-compatible modifications.

Sieberkev:
> # Release 7.0 # 
> Breaking backward compatibility:
>  * update checkstyle to jdk8 as runtime jdk

7.0 should work with Java JDK8, no?

Zizekftw:
It should work. Our work hasn't upgraded from 6.14 yet as there were no new features that we needed so I can't comment on that specifically.

I just know at version 6.14 those TreeWalker checks weren't working. I think those features were removed from the Eclipse plugin so it would throw an "Unable to instantiate module <moduleName>"

FWIW: One other check we made that was useful was to ensure that no println statements were included so that everything committed had logging done via the application logger. Up to you guys if you think it's valuable, but here's the xml to include if you want it.

```
<module name="Regexp">
    <metadata name="net.sf.eclipsecs.core.comment" value="System.out.println statements are not allowed."/>
    <property name="format" value="System\.out\.println"/>
    <property name="illegalPattern" value="true"/>
    <property name="ignoreComments" value="true"/>
</module>
```

Sieberkev:
Borked up my fork repository/branch, will resubmit

PR #1194, opened 2016-08-21
Added REST API features for Poppo UI.

JoeyyT:
Will update this more often.


Sieberkev:
You do not need to merge the whole HTML/CSS/JS/... code to the `gh-pages`, you can add your existing repository to the `.gitmodules` file so it will make a directory/link to it (only the readme and index.html file needs to be changed to link to that directory).

So if you want to add/update stuff on your GUI, you can do so on your own repository without having to wait on PR merges here :)

TTL55/scripts/app.js
```diff
@@ -0,0 +1,618 @@
...
+	$(".pokemonListDiv").show();
+	$(".editPokemon").hide();
+	$("#close").show();
+	$("#return").hide();
+}
+
+
+
+function getUrl() {
+	return "http://localhost:8080/api/bot/default/";
```
Peyphour: Here, if my bot is somewhere else or has a different name ?

JoeyyT: Yeah outdated getURL(); its fixed
One sec

Peyphour:
Is it possible de fetch the bot names from `/api/bots` instead of asking the user to input them ?

Sieberkev:
Merged as a git submodule in #1215

PR #1197, opened 2016-08-21
Lowest km egg in infinite incubator

kom-au:
New to github, not quite sure how to do things yet... This is something I do in game otherwise you'll eventually end up having to put 2km eggs in disposables. Please disregard if I've not done the contribution properly or it's not something that people want by default.

**Changes made:**
* Puts lowest km egg in infinite incubator

### DIFF
--- a/src/main/kotlin/ink/abb/pogo/scraper/tasks/HatchEggs.kt
+++ b/src/main/kotlin/ink/abb/pogo/scraper/tasks/HatchEggs.kt
@@ -45,9 +45,13 @@ class HatchEggs : Task {
                 .filter { !it.isIncubate }
                 .sortedByDescending { it.eggKmWalkedTarget }
         if (freeIncubators.isNotEmpty() && filteredEggs.isNotEmpty() && settings.autoFillIncubator) {
-            val incubateResult = filteredEggs.first().incubate(freeIncubators.first())
+			var eggResult = filteredEggs.first() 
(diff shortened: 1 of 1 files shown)


Sieberkev:
Isn't something like this already done by the sorting in #1176?

kom-au:
I'm still new to github but I think what you are referencing is a change which just sorts the incubators by their uses remaining when placing eggs inside. So if the infinite incubator is the only empty one then it'll still get the highest km egg put in it and even if a disposable incubator is free you'll still get the 2nd highest km egg put in the infinite.

src/main/kotlin/ink/abb/pogo/scraper/tasks/HatchEggs.kt
```diff
@@ -45,9 +45,13 @@ class HatchEggs : Task {
                 .filter { !it.isIncubate }
                 .sortedByDescending { it.eggKmWalkedTarget }
         if (freeIncubators.isNotEmpty() && filteredEggs.isNotEmpty() && settings.autoFillIncubator) {
-            val incubateResult = filteredEggs.first().incubate(freeIncubators.first())
+			var eggResult = filteredEggs.first() 
+			if(freeIncubators.first().usesRemaining == 0) {
```
FrancYescO: are you sure that infinite incubator got usesRemaining == 0 ?

PR #1233, opened 2016-08-25
Fix CORS for REST api

FrancYescO:


gise88:
It works.. Great job! Please merge it ASAP

Sieberkev:
There is already a `@CrossOrigin` annotation enabled on controller level, see also https://spring.io/blog/2015/06/08/cors-support-in-spring-framework

gise88:
Yes, you are right but without those @CrossOrigin I can't connect to http://ui.pogobot.club/TTL55/ correctly..

Sieberkev:
What browser are you using? In the beginning it didn't work at all in a normal browser, but after adding the annotation to the controller it works in Chrome.

Sieberkev:
I've checked the TTL55 console output a bit because I was really curious about this strange behaviour.

It looks like TTL55 does not fetch the access token correctly (uses probably a normal POST body instead of a RAW POST body, this adds a "=" to the api_rest_password) because internally it saves the access token as "Unautorized", and uses this as a token to send in the X-PGB-ACCESS-TOKEN header:

image

This probably breaks the final request after the preflight check (which returns with a status 200 instead of 50x).

gise88:
I have tried on firefox 48.0.1 and chrome 52.0.2743.116 (64-bit) in archlinux system and no way to get it works.. 
After making this change is finally gone well! I don't know.. It may be something else, but I did not really ideas of what! :smiley: 
I'll try again in the afternoon without all these `@CrossOrigin` annotation! I will let you know!

Sieberkev:
We tried it with several people on the discord channel without problems, even @FrancYescO got it working on some point.

The only way I can recreate this with:
- Start a fresh bot
- Open GUI and do a valid authentication (a bad auth will cause this problem too but that is "working as intended" since your auth is invalid, the CORS gates are closed) => GUI works flawless (getting eggs/inventory/close bot/...).
- Open another window/browser and fill in the auth form
- The internal auth token gets regenerated by this, the token used in the first window will be invalid and the CORS gates get closed on that one, generating the errors.

gise88:
I don't know if it is due from these `@CrossOrigin` annotations but running from the latest commit (https://github.com/jabbink/PokemonGoBot/commit/9c107865387ab0e96f33bac308742ca94fb503c1) I am getting these stack traces in the log file ( http://pastebin.com/QsVWm5QJ ) and it get 500 error in each API request. And yes, the password is correct infact the Auth request return a token (so, not Unauthorized)!

The same on firefox or chrome..

Sieberkev:
Do you have one (or more) JSON files in the bot-settings directory (which are not named default.json)? If so, replace the "default" in the GUI with the new botname(.json) name.

gise88:
I had already checked it, and no! Only one named default.json.. :cry:

Sieberkev:
Is this a `./gradlew run`?

gise88:
nope: `./grandle clean build ; java -jar PokemonGoBot-0.6.0.jar`

PR #1263, opened 2016-08-31
Slashes missing in API.md

FrancYescO:


Sieberkev:
O_o I've never noticed that

Issue #1139, opened 2016-08-17
Cannot understand API login process


MaxFrax:
i'm not able to understand how to login to rest APIs.
The required parameter {name} is my player's nick name?
Should I post as body this raw string? `rest_api_password`
If I try this combination I get 400 error as shown below
![immagine](https://cloud.githubusercontent.com/assets/3145452/17735888/13eb9706-6485-11e6-90fc-850b53d04702.png)

(i'm using postman)

jabbink:
Hi @MaxFrax,

It looks like there is some information missing from your ticket that will be needed in order to diagnose and fix the problem at hand. Please take a look at the [Contribution Guidelines](https://github.com/jabbink/PokemonGoBot/blob/develop/.github/CONTRIBUTING.md) and [default Issue template (the one you did not follow)](https://github.com/jabbink/PokemonGoBot/blob/develop/.github/ISSUE_TEMPLATE.md), which will tell you **exactly** what your ticket has to contain in order to be processable.

I'm marking this one now as needing some more information. Please understand that if you do not provide that information within the next two days (until 2016-08-19 14:20) I'll close this ticket so it doesn't clutter the bug tracker.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

JoeyyT:
If using config.properties use: localhost:8001/api/bot/default/auth
Then u should get an auth key :)

Sieberkev:
8001 is the socket port per bot (to fetch the most recent data), you need to use port 8080.
The {name} is the name of your bot (= `default` by default).
The `rest_api_password` is a random value you have to configure in your config.properties file (see config.properties.template if you don't have it already).

Firstly you need to do the auth with in the raw body the random value of `rest_api_password`, then you will get a random token back. Use that token in the `X-PGB-ACCESS-TOKEN` header in all other calls.

MaxFrax:
@Sieberkev thank you! It works!
I think updating API.md with this informations would be a good idea, do you agree?

MaxFrax:
<img width="949" alt="schermata 2016-08-17 alle 16 34 25" src="https://cloud.githubusercontent.com/assets/3145452/17740348/98f5bcca-6498-11e6-9573-5beb06d879d7.png">

Well I'm having another issue, I get 200 when I try any get request but I don't receive any content

Sieberkev:
[...]

PR #1264, opened 2016-08-31
[REST API] Added HTTP status codes for bad auth/requests

Sieberkev:
**Fixed issue:** https://github.com/jabbink/PokemonGoBot/issues/1139#issuecomment-241186713

**Changes made:**

* Added status codes on [bad requests 400](http://stackoverflow.com/questions/3290182/rest-http-status-codes-for-failed-validation-or-invalid-duplicate) (`the server cannot or will not process the request due to something that is perceived to be a client error`)/ auth 401
* Updated API.md file


Sieberkev:
Pinging @JoeyyT to see if this has any negative consequences on your GUI (it shouldn't, with the codes it should be even be better to catch errors any errors (when the statuscode is not 200)).

API.md
```diff
@@ -1,39 +1,58 @@
...
   - POST `/api/bot/{name}/start` => Start bot
 
-* Pokemon end-point :
+* Pokemon end-point:
   - GET `/api/bot/{name}/pokemons` => List all pokemons
   - POST `/api/bot/{name}/pokemon/{id}/transfer` => Transfer pokemon
-  - POST `/api/bot/{name}/pokemon/{id}/evolve` => Evolve pokemon
-  - POST `/api/bot/{name}/pokemon/{id}/powerup` => Power-up pokemon
+  - POST `/api/bot/{name}/pokemon/{id}/evolve` => Evolve pokemon (HTTP status code 400 (Bad Request) when not enough candy)
+  - POST `/api/bot/{name}/pokemon/{id}/powerup` => Power-up pokemon (HTTP status code 400 (Bad Request) when not enough candy)
```
FrancYescO: ..or stardust

FrancYescO:
nothing will change for TTL5, rather will be better, cause it will show a failed message if not a 200 response received.

API.md
Sieberkev: I've added a stardust check (that was missing).

src/main/kotlin/ink/abb/pogo/scraper/controllers/BotController.kt
```diff
@@ -234,25 +242,31 @@ class BotController {
...
+            httpResponse: HttpServletResponse
+    ): String {
         val itemBag = service.getBotContext(name).api.inventories.itemBag
         val count = itemBag.items.find { it.itemId == ItemIdOuterClass.ItemId.ITEM_INCENSE_ORDINARY }?.count
+
         if (count == 0) {
+            httpResponse.status = HttpServletResponse.SC_BAD_REQUEST
             return "Not enough incense"
         } else {
             itemBag.useIncense()
```
FrancYescO: log if usage successfull? maybe also for luckyegg and location change request

PR #1335, opened 2017-03-04
Honor transfer & evolve delay settings and gen II items

hrk:
**Changes made:**

* Respect evolveTimeDelay
* Respect autotransferTimeDelay
* [cosmetic] display level of encountered/caught pokemon
* Gen. II item grouping (berries and evolution items)


hrk:
This requires an update PokemonGoAPI jar too...

jabbink:
And as that one isn't updated I (and anybody else) can't really test this :/

